PR #7, opened 2015-10-10
howver -> however

benmarwick:
Thanks for making this open access. I look forward to seeing the rest of the chapters!


hadley:
Thanks!

PR #8, opened 2015-11-07
Minor fixes to string manipulation chapter

Robinlovelace:


strings.Rmd
```diff
@@ -57,17 +57,17 @@ x
...
 
-Instead we'll use functions from stringr. These have more evocative names, and all start with `str_`:
+Instead we'll use functions from stringr. These have more intuitive names, and all start with `str_`:
 
 ```{r}
 str_length(NA)
 ```
 
-The common `str_` prefix is particularly useful if you use RStudio, because typing `str_` trigger autocomplete, so you can easily see all of the stringr functions.
+The common `str_` prefix is particularly useful if you use RStudio, because typing `str_` will trigger autocomplete, allowing you to see all stringr functions.
```
Robinlovelace: A figure here could also be useful, e.g: 
selection_030

hadley:
Thanks, and good idea - I'll add a screen shot

PR #10, opened 2015-11-09
Fix for string subsetting example

dgromer:
Words in example vector `x <- c("apple", "banana", "pear")` should start with an uppercase letter so the `str_to_lower` example makes sense.


hadley:
Thanks!

PR #11, opened 2015-11-18
Typo correction in file expressing-yourself.Rmd

shoili:
The discussion of the code in lines 236-243 was a little confusing with x and y so I proposed changing it to a and b. Not sure if that was just an error that crept in while rewriting and fiddling around with the sentence or a conscious decision from you.
Just corrected a couple of obvious typos apart from that.


hadley:
Thanks!

PR #12, opened 2015-12-01
Change `*` to `$` in Anchors Section of strings.Rmd

bgreenwell:
I'm just learning regular expressions, but I think you meant to use $ instead of * in the second bullet point in the section titled Anchors in strings.Rmd.


hadley:
Ooops - thanks!

PR #14, opened 2015-12-11
Update lists.Rmd

harrismcgehee:
added "for" in the walk() description.

I apologize now if convention for github works differently than this. I saw a potential typo and figured it couldn't hurt to offer the suggestion.


hadley:
Thanks!

PR #15, opened 2015-12-12
Update intro.Rmd

harrismcgehee:
Took out a few extra words; added one or two clarifying words.


hadley:
Thanks - I think I fixed these before I noticed this PR.

PR #23, opened 2016-01-07
Alternative to productionize

ashander:
In response to twitter query https://twitter.com/hadleywickham/status/685146441535209472


hadley:
Thanks - I'll use my own language when I rewrite this section.

PR #30, opened 2016-01-14
Changing s to z in Heading1 to maintain uniformity

shoili:


hadley:
I think we'll do this all at once at the end.

shoili:
Yeah, that makes sense.

PR #34, opened 2016-01-22
caption of list subset image

shoili:
the caption on the bottom right of the image describing list subsetting should be a[[4]][[1]]. It is currently the same as the bottom middle which gives a list output whereas this one unlists it.


hadley:
Thanks!

PR #36, opened 2016-01-27
Update import.Rmd

radugrosu:
typos; 
Regarding the "..tbl_dfs never do partial matching. They’ll throw an error if the column doesn’t exist", I assume this behaviour is planned for future versions, since currently (0.4.3) they do partial matching.


hadley:
Yeah, it'll be in dplyr 0.5.0

PR #46, opened 2016-02-15
Update README.md

csgillespie:
All of the associated R packages for the book can be installed trivially, since the book is a (sort of) fake package thanks to the travis-CI implementation.


hadley:
Thanks Colin!

PR #49, opened 2016-02-16
k-fold or Monte Carlo cross-validation?

bgreenwell:
I think you meant k-fold cross-validation, but the technique used in this section is more in line with Monte Carlo cross-validation (i.e., split the data randomly into train/test samples many times and aggregate the results). In k-fold, however, the data are divided (at random) into k samples of roughly equal size. So I would either change the phrase k-fold to Monte Carlo (as in my commit) to reflect what the code is doing, or change the code (probably just the partition function) to reflect k-fold cross-validation.

### DIFF
--- a/model-assess.Rmd
+++ b/model-assess.Rmd
@@ -27,7 +27,7 @@ i.e. how do dplyr and purrr intersect.
 
 ## Multiple models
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks! Have incorporated in latest work

PR #53, opened 2016-02-20
Fix typos in robust-code.Rmd

neerajt:
Found and fixed a few typos.

Where can I make comments about potentially restructuring or explaining things a little more in certain sections?


hadley:
Looks like this conflicts with some other PRs, but I wouldn't worry too much about this chapter. I've decided that it's probably too advanced for this book, and will probably go in "Intermediate R". Otherwise, opening issues is a fine way to request clarifications or suggest restructurings.

PR #60, opened 2016-03-04
typos and similar

jennybc:
I can't seem to knit this due to bookdown issues but there are small changes.


hadley:
Thanks! You probably need to install "new" bookdown with `install_github("rstudio/bookdown")`

PR #61, opened 2016-03-07
Update README

jimhester:
There was a broken link and all the R packages are installed automatically using `Remotes`.

I believe the only external dependency is now pandoc.


hadley:
Thanks!

PR #64, opened 2016-03-10
Functions.Rmd: some proposed words on identical()

ijlyttle:
Adds discussion of `identical()` function to `if` section. 

I, personally, find identical() to be useful, but I can see that it does not add aesthetic value to the code. 

(BTW, really enjoying seeing this book come together!)


hadley:
Thanks! I think I need to add something to the effect that `==` is vectorised. Maybe a whole section on testing for equality would be appropriate here

PR #65, opened 2016-03-13
fixed some typos in modeling chapter

dutri001:


hadley:
Looks like many these already got fixed. Thanks!

PR #73, opened 2016-03-25
Update iteration.Rmd

darrkj:
the loops vs FP sounded like apples to oranges, moved one layer up in abstraction


hadley:
I like it- thanks!

PR #74, opened 2016-03-25
link fixed

Nowosad:


hadley:
Thanks!

Nowosad:
You' re welcome. However, the problem still occurs. In subsection http://r4ds.had.co.nz/functions.html#when-should-you-write-a-function, there is a linking problem. Reference to the iteration section (http://r4ds.had.co.nz/iteration.html#iteration) works fine, but the link to the data structure section is not visible ([data-structures]).

PR #76, opened 2016-03-25
changed "formally" to "formerly"

kdpsingh:
I think you meant formerly (meaning "previously") rather than formally.


hadley:
Doh - thanks!

PR #88, opened 2016-04-05
tbl_dfs now do partial matching?

KyleHumphrey:
Do tbl_dfs now do partial matching? It works for me and it seems to work on the live r4ds website (immediatley above http://r4ds.had.co.nz/transform.html#dplyr-verbs). At any rate, it doesn't throw an error, as was meant to be illustrated.


KyleHumphrey:
Perhaps this should be added as an issue?

PR #93, opened 2016-05-01
Missing the word "compared"

croach:


hadley:
I don't think that word is necessary in British English. It certainly reads fine to me.

PR #94, opened 2016-05-09
Fixed typo

bgreenwell:
Or perhaps you meant to include a word after very (e.g., very strong, etc.).


hadley:
Thanks, but model-vis is currently being heavily re-written so this is likely to conflict with my changes.

PR #97, opened 2016-05-15
Update model-vis.Rmd

gsimpso4:
corrected a typo predicton


hadley:
Thanks - but model-vis is currently being heavily re-written so this is likely to conflict with my changes.

PR #98, opened 2016-05-20
Fixed typo - "saem" to "same"

kdpsingh:


hadley:
I think Garrett fixed this in the meantime - thanks though!

PR #101, opened 2016-05-20
Minor grammar fix -- removed "type of"

kdpsingh:
There was a missing word after "type of" -- potentially you meant "type of assignment." I removed "type of" to make it more concise.


hadley:
Thanks!

PR #106, opened 2016-06-03
changed instances of heirarchical into hierachical in variation.Rmd

rlzijdeman:
Thank you for the effort of composing this book. A bit cocky as non-native english speaker, but I believe 'heirarchical' is spelled 'hierarchical'. Best, Richard


hadley:
Thanks!

PR #108, opened 2016-06-05
Update relational-data.Rmd

sibusiso16:
Hi Hadley. Thanks for putting this book together. I don't believe these a data frame called carrier and the example here is joining flights2 and airlines by the key carrier.


hadley:
Thanks!

PR #109, opened 2016-06-05
Line 61: changed "as.data_frame" to "as_data_frame"

tteo:
Hullo Hadley, big fan. This is my first time forking a project on Github! In any case, I think there is a typo on line 61 here: instead of a dot, it should be an underscore.

Also, added a note on line 36 to remind the reader to install nycflights13.

Terence


hadley:
Unfortunately this was written so your change no longer applies. But thanks anyway!

PR #112, opened 2016-06-06
Update datetimes.Rmd

sibusiso16:
Hi Hadley. The flights that departed between 06:00 and 10:00 arrived early compare to flights that departed between 00:00 and 05:00.


hadley:
Thanks!

PR #120, opened 2016-06-13
Update visualize.Rmd

sibusiso16:
Hi Hadley. I changed "If you map the fill aesthetic to a third variable" to "If you map the fill aesthetic to another variable". I could be wrong but i believe this was a typo.


hadley:
Thanks!

PR #124, opened 2016-06-17
Update relational-data.Rmd - typo

koalabearski:
Small typo in the end of a paragraph on relations:

from: each airport flies to many airlines; each airport hosts many airlines.
to: each airline flies to many airports; each airport hosts many airlines.


hadley:
Ooops, thanks!

PR #125, opened 2016-06-18
Fix a typo: Should be `purrr` instead of `purr`

jjchern:
Also I guess one natural solution to this exercise would be 
```R
purrr::map_if(heights, is.numeric, ~ cor(.x, heights$income))
``` 
But it leads to some crazy outputs. And `cor` is missing a `na.rm` argument.

And I believe the `heights$id` variable should be either a string or a factor.

### DIFF
--- a/model-basics.Rmd
+++ b/model-basics.Rmd
@@ -155,7 +155,7 @@ summary(h)
 ### Exercises
 
 1.  What variables in `heights` do you expect to be most highly correlated with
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!  I don't think I've introduced `map_if` in the text, so you could also solve it in two steps, first using `map_lgl` to find the numeric columns.  

BTW `cor()` has the `use` argument instead of `na.rm`

PR #131, opened 2016-07-05
Update tidy.Rmd

sibusiso16:
The table4 and table5 do not have a column called 2001 and this line cases <- c(table4$1999, table4$2000, table4$2001) returns this error "Error: unexpected numeric constant in "cases <- c(table4$1999" and this code seems to return the expected results "cases <- c(table4$`1999`, table4$`2000`)"


hadley:
Fixed in another pull - thanks!

PR #136, opened 2016-07-08
Corrected two typos

sco-lo-digital:
Small typos ;

Genearlised --> Generalised

and 

forrests --> Forest


hadley:
This is all in the process of being completely rewritten, but thanks for the fixes!

PR #142, opened 2016-07-09
Added stopifnot() after stop()

mustafaascha:
This might complicate things, but also might be helpful to some readers. If it is helpful, but the example is too complicated, maybe add brief mention of `stopifnot(next_section == helpful)` to help intuit the purpose of stopifnot().


mustafaascha:
Also, I assign the copyright of this contribution to Hadley Wickham

hadley:
Thanks for the idea - I've integrated it by hand with a few tweaks.

PR #153, opened 2016-07-16
Correct grammatical typos in transform.Rmd

pkq:
Edit filtering and grouping exercises for grammatical consistency.

### DIFF
--- a/transform.Rmd
+++ b/transform.Rmd
@@ -226,15 +226,15 @@ filter(df, is.na(x) | x > 1)
 
 ### Exercises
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

By the way a and are are homonyms in NZ English which is why I get them confused more often than that you might expect.

pkq:
Thank you! I can't say enough about how much your packages have improved my data management workflow, and this book has really increased my understanding of them. Keep up the great work! 

Interesting tidbit about NZ English homonyms. Totally makes sense.

PR #162, opened 2016-07-22
Move "only" closer to "simulated"

wibeasley:
trivial change: To reduce the slim chance the reader initially thinks "only" is modifying "uses".


hadley:
Thanks!

PR #181, opened 2016-07-23
Update visualize.Rmd

sibusiso16:
changed "visualize your data ggplot2" to "visualize your data with ggplot2".

also made a changed "If you run this code and get the error there is no package called ‘ggplot2’" to "If you run this code and get an error there is no package called ‘ggplot2’ on your machine". I'm not sure if "machine" is the correct word to use.


hadley:
Thanks for pointing out the confusion - I tweaked the wording to better reflect that text is the message that you will see.

PR #182, opened 2016-07-23
Update visualize.Rmd

sibusiso16:
changed "This defines how variables in your dataset and mapped to..." to "This defines how variables in your dataset are mapped to..."


hadley:
Thanks! You don't need to describe the change since I can see it in the diff

PR #186, opened 2016-07-24
Missing piece of text

OaCantona:


hadley:
For future reference, this would be better as an issue.

PR #189, opened 2016-07-25
Minor Typos

tteo:
Also changed spellings to British English for consistency.

FYI. Table number of light speed measurements currently reads (#tab:variation). The code appears to be correct.


hadley:
Thanks! Although I think the whole book will get converted to American when I turn it over to O'Reilly :/

PR #194, opened 2016-07-26
fix typo

nwaff:
Single typo edit


hadley:
I always forget those nots!

PR #200, opened 2016-07-28
Change to avoid redundancy in sentence

mine-cetinkaya-rundel:
Last sentence in EDA introduction reads "We'll come back to modelling in modelling" which sounds redundant, change first modelling to "it".

Also, 38 models of "cars" should be plural.


hadley:
The challenging thing with the links is that have text on the website, but in the book will look like "Section 4.6" etc. But I think you change reads better in both cases.

Thanks!

PR #202, opened 2016-07-28
Spelling change for consistency in Welcome

iansealy:
Standardising visualizing -> visualising as the latter seems to be used more often.


hadley:
Thanks!

PR #207, opened 2016-07-29
Few typos in EDA, transform, and visualize

mine-cetinkaya-rundel:
Mostly text typos, one commit touches the code moving the mapping = aes() into the geom layer to match other ggplot calls in that section.


hadley:
Thanks - this is great!

PR #209, opened 2016-07-30
Typos in transform & EDA

mine-cetinkaya-rundel:
Typos in transform and EDA + adding mapping before aes in ggplot calls until the last section of EDA where you discuss you'll be omitting them from that point onwards.


hadley:
Thanks!!

PR #210, opened 2016-07-31
Fixes in tidy

chrMongeau:
Some typos are fixed and "http" protocol is added to a link so that it can be displayed.


hadley:
Thanks!

PR #224, opened 2016-08-02
Fixed a reference to nchar(NA) and added examples

chrMongeau:
- The reference to nchar(NA) giving 2 as output was removed (as it's not actual). The "str_*" autocompletion was moved at the beginning.
- Added some examples to the "Repetition" subsection .


hadley:
Thanks - this is great! I'm integrated by hand because I'm currently working on the strings chapter.

PR #225, opened 2016-08-02
some corrections on chapters 19 and 20

tonyfischetti:
I just read over chapters 19 and 20. Here are some minor corrections.

Just one note on the changes I *did* make: I 'fixed' "so called" to be consistent with "so-called" which is used elsewhere in the chapter

Here are some potential things I found but wasn't confident enough to put in the pull request
- "guaranteed to find one local optima." I think optima is the plural. If it's only one, I think it's "optimum". Maybe you know something I don't about that
- In the beginning of section 20.6, it says "Pattern in variance, not just mean. Or other characteristics of distribution of residuals." That looks like a NOTE to me, is that supposed to be there?

### DIFF
--- a/model-basics.Rmd
+++ b/model-basics.Rmd
@@ -186,7 +186,7 @@ ggplot(sim1, aes(x, y)) +
   geom_abline(intercept = best$par[1], slope = best$par[2])
 ```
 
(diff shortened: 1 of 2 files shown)


tonyfischetti:
There are still some errors in the pull request that were not fixed by 33d7e6e

PR #230, opened 2016-08-04
Update datetimes.Rmd - Fix double and/and

csrvermaak:
Fix double and/and


hadley:
Thanks! BTW you don't need to describe the change because I can see it in the diff.

PR #241, opened 2016-08-06
Update strings.Rmd

sibusiso16:
not sure which to pick "What patterns does this regular expression match" or "What patterns will this regular expression match"


hadley:
I've fixed this locally: `` What patterns will the regular expression `"\..\..\..` match?  ``

Thanks!

PR #242, opened 2016-08-09
Not applicables -> Not availables

etiennebr:
I think it's more accurate. ref.: (https://stat.ethz.ch/R-manual/R-devel/library/base/html/NA.html)


hadley:
Thanks!

PR #243, opened 2016-08-09
keep mpg dataset

etiennebr:
I think the concept is easier to understand by staying consistent with the `mpg` dataset, there's probably no need to introduce a new dataset for that remark only.


hadley:
Thanks!

PR #267, opened 2016-08-15
Test rather than confirm hypotheses

jonpage:
While "confirmatory analysis" is a thing, focusing on hypothesis testing rather than hypothesis confirmation fits better with the mention of falsifiability.


hadley:
I think changing it to hypothesis testing makes it feel too narrow

PR #275, opened 2016-08-17
Update communicate-plots.Rmd

sibusiso16:
typo


hadley:
Thanks - merged by hand since they conflicted with some other changes.

PR #279, opened 2016-08-17
Trivial typo in fun name

gavinsimpson:
Typo: `gsave()` -> `ggsave()`

### DIFF
--- a/communicate-plots.Rmd
+++ b/communicate-plots.Rmd
@@ -503,7 +503,7 @@ It's also possible to control individual components of each theme, like the size
 
 ## Saving your plots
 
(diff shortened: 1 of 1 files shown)


hadley:
Already fixed myself :/

gavinsimpson:
Oops, sorry - never occurred to me to check the source but I guess there is a delay before the website is updated. I'll check next time before doing a PR.

(It's a bit of a problem with the edit button in Bookdown sites though... if a user has to go check the actual source anyway...)

hadley:
I don't think there's any way to avoid in this case - the edits were minutes apart

PR #287, opened 2016-08-19
fix code formatting in communicate-plots

tklebel:


communicate-plots.Rmd
```diff
@@ -535,7 +535,7 @@ I only ever use three of the five options:
 
 * To put multiple plots in a single row I set the `out.width` to
   `50%` for two plots, `33%` for 3 plots, or `25%` to 4 plots, and set
-  fig.align = "default"`. Depending on what I'm trying to illustrate (e.g.
+  `fig.align = "default"`. Depending on what I'm trying to illustrate (e.g.
```
tklebel: Is there a reason why you use double quotes here (`fig.align = "default"`), but in line 533 single quotes (`fig.align = 'center'`)?

hadley: No, just a think-o

PR #310, opened 2016-08-26
Adding links to python and Julia

csgillespie:


hadley:
I don't think they that helpful and will be a bit distracting in the printed text since they'll need footnotes or similar

PR #316, opened 2016-08-26
Formatting object_name

csgillespie:


hadley:
I wonder if it should be "object name gets value" since you the underscore is silent ;)

PR #325, opened 2016-08-26
Describe chunk structure

jennybc:
I'm not sure if this is how you'd word this but something wasn't quite right.


hadley:
Thanks!

PR #334, opened 2016-08-29
intro - minor syntax fixes

seanpwilliams:
the largest change was adding “\” to get the twitter handles to render
properly


hadley:
Thanks!

PR #411, opened 2016-09-07
Changing frame_data to tribble

csgillespie:


csgillespie:
Just realised that `tibble` hasn't loaded. It's probably easier if you close this pull close and make the changes you want. Also I just noticed you've a few `data_frame`'s kicking about.

PR #434, opened 2016-09-22
fix typo in ch 30 R Markdown workflow

benmarwick:
I see you're not mentioning git in this book at all, but on the topic of good habits for writing Rmd files, I have found it useful to have a colophon code chunk at the end of a long Rmd that outputs git repo and commit details, like this:

---
### Colophon

This report was generated on `r Sys.time()` using the following computational environment and dependencies: 

```{r colophon, cache = FALSE}
# which R packages and versions?
devtools::session_info()
 
# what commit is this file at?
library(git2r)
repo <- repository(path = "../..")
last_commit <- commits(repo)[[1]]
```

The current git commit of this file is `r last_commit@sha`, which is on the `r branches(repo)[[1]]@name` branch and was made by `r last_commit@committer@name` on `r when(last_commit)`. The current commit message is "`r last_commit@summary`". The repository is online at https://github.com/benmarwick/xxx

---

With this bit of text, I can easily keep track of what my co-authors are looking at when I email new versions of the output to co-authors (who are not using R or git), as we iterate over a project (which can take years).  

I know it's a bit less than ideal to use the slot accessor `@` like that, any tips to avoid that?

### DIFF
--- a/rmarkdown-workflow.Rmd
+++ b/rmarkdown-workflow.Rmd
@@ -55,7 +55,7 @@ Much of the good advice about using lab notebooks effectively can also be transl
     __packrat__, <http://rstudio.github.io/packrat/>, which store packages 
     in your project directory, or __checkpoint__,
     <https://github.com/RevolutionAnalytics/checkpoint>, which will reinstall
-    packages avaialble on a specified date. A quick and dirty hack is to include
+    packages available on a specified date. A quick and dirty hack is to include
     a chunk that runs `sessionInfo()` --- that won't you let easily recreate 
     your packages as they are today, but at least you'll know what they were.
 


hadley:
@jennybc maybe a function that wrapped up some of this would be useful for githug? Then it would just be a single command to get all the git details.

jennybc:
Thanks for the suggestion! (Too bad there's no option to view "raw" of issue comments.)

benmarwick:
@hadley @jennybc thanks for taking a look, and the pointer to githug!

hadley:
I like the idea, but I feel like it doesn't fit quite right at the moment, and the book needs to be done ASAP. Will definitely think about this again for the 2nd edition!

benmarwick:
No worries, thanks for taking a look. I see you got the typo fixed in another PR, so nothing more to do here.

PR #437, opened 2016-09-24
Fix typos

behrman:


hadley:
I think something is slightly buggy with your github workflow because you seem to be generating pull requests that include the contents of previous pull requests (e..g. this one contained #435 and #436)

behrman:
I created a branch for each chapter. Most chapters had just one commit and pull request. But for some chapters, I notice additional typos after submitting a pull request. For these I checked out the branch again, fixed the additional typos, did a commit and push. I may not need to do another pull request at this point. The push may work its way through. Is this so? The bug may be that in some cases I submitted another pull request for the updated branch.

hadley:
Yeah, a pull request always tracks a branch, so you don't need to create a new PR.  GitHub does a really bad job of surfacing this idea, but it makes iterative development in a PR much easier.

behrman:
Yes. I think I figured this out, but while doing so submitted unnecessary PRs.

PR #440, opened 2016-09-26
Typo fix

alexpghayes:
In model basics: Typo fix: algorthim to algorithm. Removal of possible stub line after Other model families section header. New to github - hopefully I'm doing this right.


hadley:
Thanks! Had to merge by hand because someelse pointed out the same algorithm typo

PR #447, opened 2016-09-30
Rogue speech marks

csgillespie:


hadley:
Thanks - someelse else got it too

PR #449, opened 2016-10-02
Fix variable name mismatch

jpetuchovas:
Variable name is "relig" not "religion".


hadley:
Thanks, also caught by someone else.

PR #456, opened 2016-10-06
Fixing typos

gridgrad:
1- 
"The following code fits a model that predicts price from carat and the**n** computes the residuals (the difference between the predicted value and the actual value)." 

2-

"You’ll learn how models, and the modelr package, work in the final part of the book, [model]**(#model-intro)**. We’re saving modelling for later because understanding what models are and how they work is easiest once you have tools of data wrangling and programming in hand."


hadley:
Thanks!

PR #460, opened 2016-10-08
Update visualize.Rmd

sibusiso16:
In Workflow: basics, you have TAB and ESCAPE in caps and made a change here to be similar.


hadley:
Thanks!

PR #462, opened 2016-10-10
fix typos

yahwes:
import.Rmd: 

- "But it you many" -> "But **_if_** you **_have_** many"

program.Rmd:

 - "That means getting better at programming also involves get better at communicating." -> "That means getting better at programming also involves **_getting_** better at communicating."


hadley:
Thanks! Don't worry about explaining simple changes like this since I can easily see them in the diff.

PR #466, opened 2016-10-12
Update workflow-basics.Rmd

Nowosad:


hadley:
Thanks, although I think this might be a UK/US English thing, and I might have been inconsistent throughout the book.

Nowosad:
I find lower case "it" after a colon in a few places in the book, so maybe it's good thing to unify that.

PR #470, opened 2016-10-13
Correct little typos

nchachereau:
Two typos I noticed as I was reading.

Thank you for this very good introduction!


hadley:
Thanks - I fixed it in a slightly different way

PR #479, opened 2016-10-19
Formula correction

jroberayalas:
I think there's a typo here. If we define the formula log(y) ~ sqrt(x1) + x2, then this is transformed to log(y) = a_1 + a_2 \* x1 \* sqrt(x) + a_3 \* x2.


hadley:
Both seem wrong - I think it should be `log(y) = a_1 + a_2 * sqrt(x1) + a_3 * x2`

jroberayalas:
Oh, yes, agree! I was paying more attention to the missing of the log(.) function.

PR #480, opened 2016-10-19
Typo (and redundant) correction

jroberayalas:
There's a typo in the word 'focused'. But overall, the sentence seems like a repetition of the previous one.


hadley:
Thanks - I tweaked it manually because I wanted to emphasise the key difference is simulated vs. real data.

PR #490, opened 2016-10-26
Typos in Vectors chapter

nirmalpatel:
Small typos in Vectors chapter for which I had submitted issue. Then I found out I can create pull requests!

I've also suggested small changes, because the book is up for publication.

Thanks for writing such a great book!


hadley:
Thanks!

PR #492, opened 2016-10-30
Correct small typo in intro.Rmd

BarkleyBG:
Last sentence in section 1.4.1: R should be changed to "putting it off only makes it worse" or "putting it off will only make it worse".


hadley:
Thanks!

PR #500, opened 2016-11-23
Update visualize.Rmd

kaetschap:
function used is tribble() but function used in the text below to describe the code snippet was tibble()
=> changed from tibble() => tribble()


hadley:
Thanks!

PR #503, opened 2016-11-28
correct typo

micongrin:
"make" to "makes"


hadley:
Thanks - fixed in another PR.

PR #512, opened 2016-12-19
Typo in Unusual values

RobinGertenbach:
the unusually wide limits are either on the y-variable or on the x-axis


hadley:
Thanks!

PR #514, opened 2017-01-03
added hint for cummulative functions

Adrianzo:
This excercise it's difficult. You should also consider moving it to the next section (5.7) when window functions are introduced.


hadley:
Thanks!

PR #526, opened 2017-01-22
Update Functions.Rmd

agila5:
I think there is a typo in wt_mean because if it's a weighted mean of x with weights w it should be sum(x * w) / sum(w)

(Sorry if I did something wrong, i'm a newcomer to github)


hadley:
Ooops, yes, thanks!

PR #543, opened 2017-02-24
Add `show.legend = FALSE` mentioned in exercise

djbirke:
The example wants to show different ways of producing the same plot (as done in the print version). One of the exercises refers to a `show.legend = FALSE` used previously. This change brings the parts back into harmony.


hadley:
Thanks!

PR #550, opened 2017-03-15
Two changes suggested by Toshi

garrettgman:
1. Adds a select() call when exploring diamonds sizes to ensure that the anomalous y variable appears on the page of the printed book (where it is discussed).
2. Adds message = FALSE to suppress "Loading required package: methods" message.


hadley:
Thanks! You can make these changes directly too.

PR #563, opened 2017-03-17
Vector.Rmd: Typo Corrected "Augmented Vectors"

saghirb:
There is a difference in four augmented types between Section "Augmented Vectors" and Section "Basic Vectors". Please double check.


hadley:
Thanks!

PR #568, opened 2017-03-27
correct typo (fixes #506)

### LINKED ISSUE #506: Ch 20.1 - Missing word
In the paragraph:

> Vectors are particularly important as most of the functions you will write will work with vectors. It is possible to write functions that work with tibbles (like ggplot2, dplyr, and tidyr), but the tools you need write such functions are...
You're missing a word;
but the tools you need **to** write such functions

katrinleinweber:
cc @lmadsenbc


hadley:
Thanks!

katrinleinweber:
You're welcome :-)

PR #570, opened 2017-04-03
Fix duplicate figures

talgalili:
I don't see a reason to have two "group = drv" plots.
Probably the second one was meant to be "linetype = drv" (as corrected), in order to demonstrate the difference between showing (or not) the legend.


hadley:
Thanks - fixed in another PR too.

PR #571, opened 2017-04-04
Keep the same "table of cut values" example

talgalili:
No need to go with imaginary values when there is a nice example that is used before and after this example.


hadley:
Thanks!

PR #573, opened 2017-04-14
Adds the missing word "top"

evjrob:
The following sentence under the section Vector Basics is missing the word "top": "Dates and date-times are built on of numeric vectors." The sentence should read: "Dates and date-times are built on top of numeric vectors."


hadley:
Thanks! BTW you don't need to describe the change in the text as I can see it in the diff :)

Issue #576, opened 2017-04-21
Factor Reordering section - clarify relig data frame vs. factor


bbrewington:
I would recommend calling the summary data frame (tibble?) "relig_summary" or "relig_df", to make it more clear in the ggplot call with fct_reorder which is the data frame and which is the factor variable

![image](https://cloud.githubusercontent.com/assets/10573749/25282697/42067298-267f-11e7-90c3-aaf5fe8fdb2f.png)

![image](https://cloud.githubusercontent.com/assets/10573749/25282717/512a2c60-267f-11e7-8935-aeeae925a9e4.png)

bbrewington:
I know this may seem like a moot point, but I was pretty confused by the naming at first (when I was just looking at the ggplot call, and before I ran the code myself to inspect)

hadley:
Want to do a PR?

bbrewington:
Done - https://github.com/hadley/r4ds/pull/577

PR #577, opened 2017-05-04
factors.Rmd clarification

bbrewington:
Changed the name of the "relig" data frame to "relig_summary" in order to better distinguish between the factor variable "relig" and the data frame it belongs to (This caused a bit of confusion for me as I was reading through R4DS)

Also performed the same correction for data frame "rincome"


hadley:
Thanks! For future reference, if you include `fixes #576` in the commit message it'll automatically close the issue when the PR is merged (I did it for you this time).

bbrewington:
Good to know.  I'll make a point to do that going forward

PR #582, opened 2017-05-19
Correct stated reduction in amount of code

twgardner2:
It seems to me that extracting the exponent to the argument of the new function results in 2/3 less code (1/3 of the original code).


hadley:
I think I'd prefer to word this as "1/3 of the original code". Otherwise it's a bit confusing.

PR #584, opened 2017-05-22
Use visible column

krlmlr:
for `desc()` example; the previous `arr_delay` column was invisible in the website rendering.


hadley:
Thanks!

PR #587, opened 2017-06-30
Introduce `dplyr::case_when()` as an alternative to `ifelse()`

maurolepore:
Explain when ?dplyr::case_when() if more useful than ifelse.

- Copy-pasted text from ?dplyr::case_when()


hadley:
Thanks

PR #590, opened 2017-07-08
Update model-basics.Rmd

JamesThompsonC:
change "the simulated data set `sim1`"  to
"a simulated data `set1`". The "the" in the sentence implies that sim1 has been mentioned before.


hadley:
I think the "the" implies that it's found in modelr, and hence is ok.

Issue #594, opened 2017-08-02
5.5 Add new variables with mutate()


DaLoneRider:
The code in the book is:

```
flights_sml <- select(flights, 
  year:day, 
  ends_with("delay"), 
  distance, 
  air_time
)
mutate(flights_sml,
  gain = arr_delay - dep_delay, 
  speed = distance / air_time * 60
)
```

I believe that "gain" implies reduction in the delay, therefore the code should be: 

```
flights_sml <- select(flights, 
  year:day, 
  ends_with("delay"), 
  distance, 
  air_time
)
mutate(flights_sml,
  gain = dep_delay - arr_delay,
  speed = distance / air_time * 60
)
```

The same (IMO inaccurate) approach to estimating the gain is used in the following two examples as well.

PR #595, opened 2017-08-02
Changed gain to be dep_delay - arr_delay, and not vice versa.

rudeboybert:
At risk of being a pedant who overly engages in strictly semantic debates, I believe @DaLoneRider is correct in #594. At the outset of Section [5.5 Add new variables with `mutate()`](http://r4ds.had.co.nz/transform.html#add-new-variables-with-mutate) you have the following example:
```
library(tidyverse)
library(nycflights13)
flights_sml <- select(flights, 
  year:day, 
  ends_with("delay"), 
  distance, 
  air_time
)
mutate(flights_sml,
  gain = arr_delay - dep_delay,
  speed = distance / air_time * 60
)
``` 
which yields
```
#> # A tibble: 336,776 × 9
#>    year month   day dep_delay arr_delay distance air_time  gain speed
#>   <int> <int> <int>     <dbl>     <dbl>    <dbl>    <dbl> <dbl> <dbl>
#> 1  2013     1     1         2        11     1400      227     9   370
#> 2  2013     1     1         4        20     1416      227    16   374
```
The first flight departed 2 minutes late and arrived 11 minutes late, for a loss of 9 minutes in the air. I think of "gain" as a good thing, so losing 9 minutes in their air should be a gain of `-9`, and not `9` as in the output. This pull request fixes this.

This is unless you meant "gained delay", in which case the output is correct. 😃

### DIFF
--- a/transform.Rmd
+++ b/transform.Rmd
@@ -325,7 +325,7 @@ flights_sml <- select(flights,
   air_time
 )
 mutate(flights_sml,
-  gain = arr_delay - dep_delay,
+  gain = dep_delay - arr_delay,
   speed = distance / air_time * 60
 )
 ```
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #601, opened 2017-08-21
Update vectors.Rmd

sedaghatfar:


hadley:
The existing wording seems ok to me

PR #611, opened 2017-09-21
Code output twice

Tazinho:
The comment should be deleted or the eval set to FALSE. I chose the latter.


hadley:
I think the former would be better

Issue #598, opened 2017-08-09
Small typo on runif in "Many Models"


alexhallam:
```r
sim <- tribble(
  ~f,      ~params,
  "runif", list(min = -1, max = -1),
  "rnorm", list(sd = 5),
  "rpois", list(lambda = 10)
)

sim %>%
  mutate(sims = invoke_map(f, params, n = 10))
```

I think that `max` in `runif` is supposed to be 1 not -1. Giving the following:

```r
sim <- tribble(
  ~f,      ~params,
  "runif", list(min = -1, max = 1),
  "rnorm", list(sd = 5),
  "rpois", list(lambda = 10)
)

sim %>%
  mutate(sims = invoke_map(f, params, n = 10))
```

PR #614, opened 2017-10-16
Update model-many.Rmd

landesbergn:
Fix typo on line 387 to use runif with min = -1 and max = 1

Addressing issue: https://github.com/hadley/r4ds/issues/598


hadley:
Thanks!

PR #615, opened 2017-10-23
Strings chapter: elaborate on capturing groups

florisvdh:
They idea of how _Grouping and backreferences_ work should be better elaborated.

After reading, it was not clear to me that `\1` etc. actually refer to the text matched by the first, ... capturing group. It was after reading some stuff on [this](https://www.regular-expressions.info/brackets.html) and [this](https://www.regular-expressions.info/backref.html) site that I understood this better. Therefore I made a suggestion to clarify this.

### DIFF
--- a/strings.Rmd
+++ b/strings.Rmd
@@ -389,7 +389,7 @@ str_view(x, 'C[LX]+?')
 
 ### Grouping and backreferences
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #616, opened 2017-10-30
Update pipes.Rmd (minor typo)

mbeveridge:
Querying whether there is a minor typo on row 130 : 'mouse' --> 'mice'
...To use the same 'up = field_mice' as all the other code examples


hadley:
Thanks!

PR #621, opened 2017-11-12
Better naming consistency

hao-trivago:
Now 'key' is used commonly across step-by-step guide, final complex pipe, and the exercise.


hadley:
Thanks!

PR #622, opened 2017-11-21
Typo in R Markdown workflow

mungojam:


mungojam:
Already in there by somebody else, sorry, didn't check first

PR #627, opened 2017-12-01
Update model-building.Rmd

mbeveridge:
24.2.3 Exercises Q3 : Another letter or word should follow 'the'. In the hardcopy book, it says 'they' (and that is the edit I've made) ..."Are they particularly bad or good"


hadley:
Thanks!

PR #636, opened 2018-01-10
Change chapter title

CamillaPhilipp:
I looked for joins and didn't find them


hadley:
I think the chapter title is ok as is - joins are an important part of relational data.

PR #640, opened 2018-02-04
Word swap

jeffboichuk:
Fixes #640 to remain consistent with this sentence: "Since we already use the word "value" to describe data, let's use the word "level" to describe aesthetic properties."


hadley:
Thanks!

PR #646, opened 2018-02-13
by_age, regroup only by age for proportions

erikerhardt:


hadley:
Thanks - was fixed earlier.

PR #650, opened 2018-02-23
added "root" to "mean squared deviation"

EdwinTh:
Since you are referring to the standard deviation here. Without root it is the variance, isn't it?


hadley:
Good catch - thanks!

PR #653, opened 2018-03-01
Fix bug in proportion

maxharp3r:
Running the code as-is results in all proportions == 1 (and therefore uninteresting charts). Changing count to summarize fixes the issue.


hadley:
Fixed by a few others - thanks!

PR #654, opened 2018-03-01
Update factors.Rmd

pebbe:
It didn't work. All values for 'prop' were 1.


hadley:
Fixed by a few others (and any use of `$` inside dplyr is a warning sign)

PR #657, opened 2018-03-09
change make_prediction() to model1()

AlanFeder:
In the second exercise, you reference a function make_prediction(), which is never previously defined.  Comparing the measure_distance() function to earlier, it seems as if you mean to reference the model1() function.
This was noticed in the r4ds slack group on March 7, 2018 9:26 PM EST


hadley:
Thanks!

PR #662, opened 2018-03-15
Update datetimes.Rmd (trivial typo)

mbeveridge:
Row 362 was : 'LARGER units are created by CONVERTING minutes, hours, days, weeks, and years TO seconds at the standard rate'.
Proposed : 'LARGER units are created by CONVERTING minutes, hours, days, weeks, and years FROM seconds at the standard rate'

I may be wrong, (and it's trivial), but keep 'bumping' on it when I read

### DIFF
--- a/datetimes.Rmd
+++ b/datetimes.Rmd
@@ -359,7 +359,7 @@ dweeks(3)
 dyears(1)
 ```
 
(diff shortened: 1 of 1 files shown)


hadley:
"To" sounds correct to my ear.

PR #674, opened 2018-04-09
Update model-many.Rmd (missing word)

mbeveridge:
row 269 : changed "It's possible avoid this join" --> "It's possible to avoid this join"


hadley:
Thanks!

PR #678, opened 2018-05-14
typo correction in "Durations"

jenren:
From: "Because of DST, March 12 only has 23 hours, so if add a full days worth of seconds we end up with a different time."
To: "Because of DST, March 12 only has 23 hours, so if we add a full days worth of seconds we end up with a different time."


hadley:
Thanks!

PR #687, opened 2018-06-20
Mention the use of a character class for metacharacters

### LINKED ISSUE #673: Single character class for regex metacharacters
Re: the chapter on strings and how to include a literal metacharacter in the string representation of the regular expression.

It might be worth mentioning that a character class with that single character is an alternative to all the backslashes for many of the metacharacters. I think the resulting regex is more readable. I wish I remembered this more often myself!

Tacking this on at the end of an existing example:

``` r
# To create the regular expression, we need \\
dot <- "\\."

# But the expression itself only contains one:
writeLines(dot)
#> \.

# And this tells R to look for an explicit .
stringr::str_view(c("abc", "a.c", "bef"), "a\\.c")

# And so does this
stringr::str_view(c("abc", "a.c", "bef"), "a[.]c")
```

Preliminary fact collecting

👍 metacharacters for which this works: `$` `.` `|` `?` `*` `+` `(` `)` `[` `{` 

👎 for which this does not work, i.e. they have special meaning inside a character class: `]` `\` `^` `-`

jennybc:
Closes #673

### DIFF
--- a/strings.Rmd
+++ b/strings.Rmd
@@ -299,6 +299,17 @@ There are a number of special patterns that match more than one character. You'v
 
 Remember, to create a regular expression containing `\d` or `\s`, you'll need to escape the `\` for the string, so you'll type `"\\d"` or `"\\s"`.
 
+A character class containing a single character is a nice alternative to backslash escapes when you want to include a single metacharacter in a regex. Many people find this more readable.
+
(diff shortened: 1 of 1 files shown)


hadley:
Perfect - thanks!

PR #691, opened 2018-07-19
Could you enable HTTPS?

yihui:
so that the bookdown.org homepage is completely secure. Currently this is the only book in the list which has not enabled HTTPS yet.

Thanks!


hadley:
Just switched to netlify, and enabled https, so I should be able to merge tomorrow.

hadley:
Thanks @yihui!

PR #707, opened 2018-08-31
Markdown format fix: Quote attribution moved into blockquote

lwjohnst86:
I assume that the attribution of the quote to the person be in the same blockquote as the quote itself.


hadley:
Yes — thanks!

PR #719, opened 2018-09-27
Removed now-redundant sentence and code segment.

MusTheDataGuy:
Removed:

> forcats is not part of the core tidyverse, so we need to load it explicitly.
> 
> library(tidyverse)
> library(forcats)

forcats is now a part of the core tidyverse and no longer needs to be loaded explicitly.

>library(tidyverse)


hadley:
Don't we still need to load tidyverse?

MusTheDataGuy:
@hadley Indeed we do; I have updated my comment.

hadley:
Fixed by another PR that also updated the string chapter — sorry it took me so long to follow up!

PR #726, opened 2018-11-11
Section 19.1 - Typo in last paragraph

FelipeSBarros:
typo fixed on:
> As well as practical advice for writing functions, this chapter also gives you some suggestions for how to style your code. Good code style is like correct punctuation. **Youcanmanagewithoutit**, but it sure makes things easier to read! As with styles of punctuation, there are many possible variations. Here we present the style we use in our code, but the most important thing is to be consistent.

To:

> As well as practical advice for writing functions, this chapter also gives you some suggestions for how to style your code. Good code style is like correct punctuation. **You can manage without it**, but it sure makes things easier to read! As with styles of punctuation, there are many possible variations. Here we present the style we use in our code, but the most important thing is to be consistent.


sedaghatfar:
@FelipeSBarros this typo looks to be intended

sedaghatfar:
No worries, 
and yes please

PR #727, opened 2018-11-12
14.4: exercise of "Detect matches" should be 4th level heading

dongzhuoer:
The solutions book even add a [pseudo section](https://jrnold.github.io/r4ds-exercise-solutions/strings.html#tools) to keep synchronized with your numbering. 

I should fire an issue to the solutions book after you approve this PR.


hadley:
Thanks!

PR #736, opened 2018-11-28
Danny Kaplan's book seems to have moved

mine-cetinkaya-rundel:
I replaced the old URL with the new one


hadley:
Fixed by another PR

PR #741, opened 2018-12-09
geom_density() as an option for geom_freqpoly()

senakim:


hadley:
I really don't like `geom_density()` because it requires much more explanation, and it can easily give confusing results in if the underlying data is not continuous.

PR #748, opened 2019-01-11
Remove use of mu in exercises

jrnold:
Two exercises use $\mu$ to mean the mean parameter of a normal distribution. However, nowhere in the book is it defined as such. For clarity replace $\mu$ with "mean".


hadley:
Good idea 😄

PR #749, opened 2019-01-11
Add equations to 19.2 Ex 4.

jrnold:
The question asks:

> Follow <http://nicercode.github.io/intro/writing-functions.html> to write your own functions to compute the variance and skew of a numeric vector.

However, the math on that page no longer renders.

Directly include the skew and variance equations in the question.


hadley:
Thank you!

PR #750, opened 2019-01-11
Rename "Practice" sections to "Exercises"

jrnold:
Some sections with exercises are titled "Practice", while most are titled "Exercises".
This commit renames all of these to "Exercises".


Fredei:
De

hadley:
Makes sense, thanks!

PR #751, opened 2019-01-12
use `stat()` instead of `..` in "3.7 Statistical transformations"

dongzhuoer:
https://r4ds.had.co.nz/data-visualisation.html#statistical-transformations


hadley:
Good catch — thank you!

PR #754, opened 2019-01-19
update tidyverse loadings

jkolacz:
Update to the tidyverse prerequsities for the strings and factors chapters. stringr and forcats now load with the core tidyverse.


hadley:
Great fix — thank you!

PR #758, opened 2019-01-23
Assuring minimal version of Rstudio

EdwinTh:
I don't think it is meant that the user should have version 1.0.0 of Rstudio, but have a version that is at least 1.0.0. The link of course refers to the latest version. A savvy reader might think he / she should start searching for 1.0.0 with the current text.


hadley:
Yeah, I think was written when 1.0.0 was something new, not something old!

PR #760, opened 2019-02-03
Revert typo

jrnold:
Issue #681 fixed a typo in the code for an exercise, `dota` to `data`. However, the exercise should have typos, since the purpose of the exercise is to fix typos.

> Tweak each of the following R commands so that they run correctly:

This PR restores the typo so the code correctly works incorrectly.


hadley:
Thank you!

PR #763, opened 2019-02-12
Corrección viñetas y sintáxis

mendivilg:


hadley:
I think this went to the wrong repo.

PR #775, opened 2019-04-16
consistent notation for out_1 and sex_female

pachamaltese:


hadley:
I wanted to match the base R conventions here since that's what you see in the model output.

PR #781, opened 2019-05-30
clear up wording in 5.3.1 exercises

### LINKED ISSUE #708: Ambiguous wording for exercises in Section 5.3.1.
A couple of the exercises in section 5.3.1 have ambiguous wording.

> Sort flights to find the fastest flights.

Does this mean fastest in terms of minimal air_time (`air_time`) or fastest in terms of average air speed (`air_time` / `distance`).

> Which flights travelled the longest? Which travelled the shortest?

I assume that this means the distance travelled (`distance`), but I could also interpret it in terms of time (`air_time`).

marwahaha:
I tried to make the wording clearer as per @jrnold 's issue.

### DIFF
--- a/transform.Rmd
+++ b/transform.Rmd
@@ -242,9 +242,9 @@ arrange(df, desc(x))
 1.  Sort `flights` to find the most delayed flights. Find the flights that
     left earliest.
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks for the clarifying language!

PR #784, opened 2019-06-13
Fixed text to correspond with code output

BinxiePeterson:
My guess is that the guess_parser() function was changed/updated since the writing of this book. From '?guess_parser', the following default is used: 'guess_integer = FALSE'. Thus column 'x' is not problematic anymore.


hadley:
Lovely - thank you for the fix!

Issue #677, opened 2018-05-10
stringr package now loaded with tidyverse


szimmer:
Chapter 14 says that stringr is not part of the core tidyverse but that seems to no longer be true.

``` r
library(tidyverse)
#> -- Attaching packages ---------------------------------------------------------------------------- tidyverse 1.2.1 --
#> v ggplot2 2.2.1.9000     v purrr   0.2.4     
#> v tibble  1.4.2          v dplyr   0.7.4     
#> v tidyr   0.8.0          v stringr 1.3.0     
#> v readr   1.1.1          v forcats 0.2.0
#> -- Conflicts ------------------------------------------------------------------------------- tidyverse_conflicts() --
#> x dplyr::filter() masks stats::filter()
#> x dplyr::lag()    masks stats::lag()
```

PR #785, opened 2019-06-14
stringr is now in tidyverse core packages. Close #783

### LINKED ISSUE #783: Update 14 Strings
In 14 Strings > 14.1.1 Prerequisites it says that "stringr is not part of the core tidyverse because you don’t always have textual data, so we need to load it explicitly."

Using tidyverse Version : 1.2.1 we have that stringr Version 1.4.0 is already attached.

eheinzen:


eheinzen:
Hmm... I don't seem to have gotten the "Close #783 " to work quite right.  Sorry about that.  Instead of trying another time, I should note that this also closes #677 .

I assign the copyright of this contribution to Hadley Wickham

eheinzen:
SIgh.  Turns out this is already fixed in #754

PR #787, opened 2019-07-02
Update import.Rmd to match current readr behavior (first 1000 NAs)

pabloedug:
I think this is the current readr behavior when the first 1000 rows contain only NAs.
Kind regards


hadley:
Thank you!

PR #788, opened 2019-07-06
Update strings.Rmd 13.5 exercise questions

microbe:
It looks that the double quotes and a backslash are missing in the 14.3.5.1 Excercise questions.


microbe:
Oops, it is 14.3.5.1 Excercise, instead of 13.5.
Sorry for the confusion.

hadley:
That's a deliberate choice to help you think about the regular expression vs. the string that represents the regular expression.

PR #795, opened 2019-07-31
name hour_perc should be hour_prop because proportion not percentage

a-rosenberg:
Small variable name change to make it more clear that the variable is a proportion not a percentage.


hadley:
Thanks!

PR #802, opened 2019-09-10
Replace `[[` with `[` on vectors.

rochelleterman:
Not sure if this was done on purpose, but I found several `[[` operators used on atomic vectors.


hadley:
I recommend using `[[` when you want to refer to a single element, reserving `[` for indexing with vectors.

Issue #807, opened 2019-10-02
str_view(), repetition at most n times yields error


vanderlindenma:
https://github.com/hadley/r4ds/blob/ce824d0a33149b34c7b1cacabf8636e132970592/strings.Rmd#L363

As I was writing lecture notes based on the "strings" chapter, I stumbled upon the exact same issue as described in this Reddit thread: https://www.reddit.com/r/rprogramming/comments/8oyy1a/stringr_str_view_repetition_at_most_n_times/

On my end, it looked like this. The intention was to show that there would be no match in this case. Instead, I get an error:

`str_view_all("Misery, Misery, Misery that’s what Lord and Miller have chosen", "(Misery, ){,3}")`

`Error in stri_locate_all_regex(string, pattern, omit_no_match = TRUE,  : 
  Error in {min,max} interval. (U_REGEX_BAD_INTERVAL)`

One of the answerers on Reddit explains:

> "`stringr` uses ICU regular expressions, which apparently do not have the `{,m}` notation - use the explicit `{n,m}`."

Since the `{,m}` "method" is presented as a viable option in the latest version of the book and I did not see an open issue related to this, I thought I'd mention it here.

Perhaps the error message when using `{,m}` is clear enough and the inclusion of `{,m}` in the book is intentional (in the sense that it discusses a valid regex, albeit one that does not work with `stingr`, and is judiciously not included in the examples of `stingr` applications)?

hadley:
This is fixed in the 2nd ed.

PR #808, opened 2019-10-02
str_view(), repetition at most n times yields error #807, and small typos in "Iteration chapter"

vanderlindenma:
Proposed fix related to this issue: https://github.com/hadley/r4ds/issues/807

A more comprehensive fix should perhaps discuss why "`{,m}`: at most m" does not work (or warn the reader that it does not)?

Or perhaps the error message when using `{,m}` is clear enough and the inclusion of `{,m}` in the book is intentional (in the sense that it discusses a valid regex, albeit one that does not work with `stingr`, and is judiciously not included in the examples of `stingr` applications)?

(The second commit corrects some small typos in the "Iteration" chapter)


hadley:
How about just changing it to `{1,m}`?

mine-cetinkaya-rundel:
Thank you!

PR #811, opened 2019-10-28
Drop na.rm=TRUE when data is already purged of NAs

pooyataher:
In transform.Rmd (Chapter 5, Data Transformation), under Counts (5.6.3), we removed na.rm = TRUE when computing mean because all NAs for arr_delay were already removed when creating not_cancelled dataset.


hadley:
Thanks!

Issue #704, opened 2018-08-23
## 5.2.2 Logical operators: Incorrect description of `filter(flights, month == 11 | 12)`


hannesgislason:
While teaching Data Science and Statistics with RStudio at the University of the Faroe Islands, I noticed this issue. 

In 5.2.2 Logical operators the online textbook (http://r4ds.had.co.nz/) claims:

## 5.2.2 Logical operators

"The order of operations doesn’t work like English. You can’t write `filter(flights, month == 11 | 12)`, which you might literally translate into “finds all flights that departed in November or December. In a numeric context (like here), `TRUE` becomes one, so this finds all flights in January, not November or December. This is quite confusing!”. 

This sentence is incorrect: ", so this finds all flights in January, not November or December. This is quite confusing!".

The code `filter(flights, month == 11 | 12)` does not only find the flights in January, since filter(flights, month == 11 | 12) does not filter flights, but simply returns flights due to the always-TRUE condition.

```{r}

# the intended filter
filter(flights, month == 11 | month == 12)

# this code does not filter the data (all months will be included)
filter(flights, month == 11 | 12)

```
best regards
Hannes Gislason

robertchu03:
I totally agree with you.  However, his explanation would make sense if he enclosed `11 | 12` in parentheses instead in the following code.

`filter(flights, month == (11 | 12))`

PR #818, opened 2020-01-02
Correct typo or explanation

robertchu03:
11 | 12 should be enclosed in parentheses otherwise the explanation is incorrect (Issue #704)


hadley:
Thanks!

PR #820, opened 2020-01-14
Updates R4ds to use pivot_ functions

garrettgman:
Updates Chapter 9. Tidy Data with tidyr to teach `pivot_longer()` and `pivot_wider()`. 
Replaces all references to `gather()` and `spread()` with `pivot_longer()` and `pivot_wider()`.

### DIFF
--- a/tidy.Rmd
+++ b/tidy.Rmd
@@ -1,3 +1,8 @@
+---
+output: html_document
+editor_options: 
+  chunk_output_type: console
+---
 # Tidy data
 
 ## Introduction
@@ -103,7 +108,7 @@ ggplot(table1, aes(year, cases)) +
 1.  Recreate the plot showing change in cases over time using `table2`
     instead of `table1`. What do you need to do first?
 
-## Spreading and gathering
+## Pivoting
 
 The principles of tidy data seem so obvious that you might wonder if you'll ever encounter a dataset that isn't tidy. Unfortunately, however, most data that you will encounter will be untidy. There are two main reasons:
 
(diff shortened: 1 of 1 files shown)


tidy.Rmd
```diff
@@ -1,3 +1,8 @@
+---
```
hadley: Can you please remove this?

tidy.Rmd
```diff
@@ -120,89 +125,93 @@ The second step is to resolve one of two common problems:
...
+  pivot_longer(c(`1999`, `2000`), names_to = "year", values_to = "cases")
 ```
 
-The columns to gather are specified with `dplyr::select()` style notation. Here there are only two columns, so we list them individually. Note that "1999" and "2000" are non-syntactic names (because they don't start with a letter) so we have to surround them in backticks. To refresh your memory of the other ways to select columns, see [select](#select).
+`year` and `cases` do not exist in `table4a` so we put their names in quotes.
 
-```{r tidy-gather, echo = FALSE, out.width = "100%", fig.cap = "Gathering `table4` into a tidy form."}
+The columns to pivot are specified with `dplyr::select()` style notation. Here there are only two columns, so we list them individually. Note that "1999" and "2000" are non-syntactic names (because they don't start with a letter) so we have to surround them in backticks. To refresh your memory of the other ways to select columns, see [select](#select).
+
+`year` and `cases` do not exist in `table4a` so we put their names in quotes.
```
hadley: Repetition of sentence above?

tidy.Rmd
garrettgman: Sure! Fixed in 27dfedf454afbae25a6d24f00c819bb1a9264a41

tidy.Rmd
garrettgman: Thanks for catching this! Removed in 27dfedf454afbae25a6d24f00c819bb1a9264a41

hadley: BTW you can just tell me you did it; you don't need to prove it since I trust you 😄

PR #822, opened 2020-01-24
Correcting a sentence!

sitapriyamoorthi:
Together, tidying and transforming are called __wrangling__, because getting your data in a form that's natural to work with often feels like a fight! 
to 
 Together, tidying and transforming are called __wrangling__, because getting your data in a form that's natural to work which often feels like a fight!


hadley:
That change doesn't read correctly to me — we're making data easy to work with.

Issue #831, opened 2020-03-04
small error in who dataset tidying


Skl8em:
Hi,

I believe there is a small typo in the code for tidying the who dataset. 
In [tidy.Rmd](https://github.com/hadley/r4ds/blob/master/tidy.Rmd) line 514 you want to separate by the newly created names_from column from line 506 and not the key column.
Else you still have names like "newrel" when you split using "_".

regards and thanks for the book!
Ben

presnell:
I believe that the correction should actually be to replace "names_from" by "key" on line 506.  The correct code actually appears later on line 548.

presnell:
Looks like the error is repeated in the 2nd exercise in this section.  Again, "names_from" should be replaced by "key" in the call to mutate().

hadley:
This has been rewritten in the 2nd edition.

PR #832, opened 2020-03-04
Update Case Study section

dykesb:
Updated `separate()` and dropping columns in Case Study.  See https://github.com/hadley/r4ds/issues/831#issue-575067260


hadley:
I think the problem was a bit upstream and was fixed by a different PR. But I really appreciate the help!

Issue #833, opened 2020-03-07
3.2.1: data frame vs. tibble


arencambre:
In Chapter 3, specifically section 3.2.1, the chapter uses "data frame" where "tibble" is more accurate.

hadley:
A tibble is a data frame. I think the fact that it's a tibble is incidental at this point in the book.

PR #834, opened 2020-03-07
Correcting uses of "data frame" to be a "tibble"

arencambre:
In chapter 3, section 3.2.1, **data frame** is used a few times to describe the **mpg** "_tibble_". This fixes that, addressing issue #833 .


hadley:
A tibble is a special type of data frame so I don't think this is quite the right fix.

PR #839, opened 2020-04-10
Update stat_summary args for ggplot2 3.3.0

cwickham:
Although the plot still works, despite using the deprecated arguments `fun.ymin`, `fun.y`, `fun.ymax`, it makes the following exercise (1.) hard for learners to complete, since passing these arguments through `geom_pointrange()` gives the error:
  ```r
Ignoring unknown parameters: fun.ymin, fun.ymax, fun.y 
```


hadley:
Thanks!

PR #851, opened 2020-05-12
Use after_stat instead of the old dotted syntax.

nischalshrestha:
The `..` syntax in 3.7 Exercises to refer to computed variable `prop` should be updated to use the best practice of `after_stat` instead. Reduces confusion and better aligns with the documentation.


hadley:
Thank you!

PR #857, opened 2020-05-25
Model-many.Rmd: map2 w/ add_residuals adds a tibble

Shurakai:
Clarify and point out that we now don't have a double column but a tibble column added by the `add_residuals` call; this is a big difference to what we did before and in my opinion worth mentioning as it is easily skipped.


hadley:
Thanks for the fix, but we're going to be dropping the modelling chapters in the second edition.

PR #859, opened 2020-05-27
Update intro.Rmd

ishan-rep:
A small correction.


hadley:
The original seems correct to me.

PR #863, opened 2020-07-09
Update visualize.Rmd

juandering:
Solid shapes are 15-20, not 15-18.


hadley:
Oops, yes, thanks for the fix!

PR #865, opened 2020-07-19
Update tidy.Rmd - Line 506 Typo

georgewang229:
Line 506: 
When fixing the string format in the `key` column using the `str_replace` function, I think the column name was mistakenly stated as "names_from", where it should actually be "key", which is the original column name.


hadley:
Good catch, thanks!

PR #870, opened 2020-09-01
change top_n() to slice_max()

rivaquiroga:
Now that `top_n()` has been superseded in favour of `slice_max()`, making the change in its only appearance in the book.


hadley:
Thanks!

PR #894, opened 2020-11-06
dplyr::count() does have a .drop option

sciencificity:
We can include levels with no observations in dplyr::count() now with the .drop = FALSE argument setting. I changed the wording and added an example. Please feel free to amend text wording, and remove example, if you wish.


hadley:
Thank you!

PR #895, opened 2020-11-07
Switch from stat() to after_stat() in chap 3.

aleloi:
The docs in of aes_eval in [most recent](https://ggplot2.tidyverse.org/reference/aes_eval.html?q=after%20_%20stat#arguments) ggplot2 3.3.2 say that after_stat() and after_scale() has replaced stat().


hadley:
Thanks!

PR #896, opened 2020-11-07
Removed reference to deprecated .drop argument

kferris10:
According to the output and documentation, `.drop` was deprecated as an argument.  I'd propose just removing this chunk and adding a comment in the original `unnest(glance)` pipeline.

Here's what is currently showing up in the R4DS book https://r4ds.had.co.nz/many-models.html#model-quality

image

Thanks for all the awesome work on this!  I found this section really helpful to read through this morning.


hadley:
Looks like a good fix, thanks!

PR #897, opened 2020-11-09
Potential typo?

sciencificity:
I don't know if it was meant to be this way, because it's actually not weird to say `data time` but assuming it was meant to be `date time` I amended:

regular data time `lubridate::as_date_time()` -> regular date time `lubridate::as_date_time()`


hadley:
Thanks!

PR #901, opened 2020-11-21
Update model-many.Rmd

zhaoxiliang:
line 220-221, the .drop is deprecated, using select to remove the three list-columns: data, model, and resids instead.


hadley:
Thanks for the fix; someone got to it before you 😄

PR #905, opened 2020-12-08
Fix typos in lists-invoke.png

j0hnchavez:
This (minor) pull request fixes typos in lists-invoke.png.

These typos are: 
- a missing `)` in `rpois(lambda = 10, n=5`
- wrong-facing quotes in `rnorm` and `runif`

P.S. I've found this book very helpful, thanks for writing it!


hadley:
These need to be fixed in the upstream omnigraffle file, so would you mind filing an issue instead?

PR #909, opened 2020-12-31
Update visualize.Rmd

Godwin-88:
Renumber the introductory exercise on ggplot2


hadley:
These are automatically numbered in the output.

PR #914, opened 2021-01-18
Tidy data edit

mine-cetinkaya-rundel:
A round of edits for the tidy data chapter, mostly on the pivoting content.

- Add motivation before pivoting: visualization to motivate `pivot_longer()` and data wrangling to motivate `pivot_wider()`
- Add an exercise on making a contingency table -- I think it's an important use case for pivoting
- Remove separating 4 digit year into 2 digit century and year since first 2 digits is not the century, highlight this functionality in a new exercise instead
- Replace pregnancy example
- Case study:
  - Use full functionality of `pivot_longer()` instead of `separate()` first (aligns better with vignette)
  - Add a final visualization to wrap it up with a more clear reward
- Minor edits/typos


tidy.Rmd
```diff
@@ -124,54 +124,120 @@ Typically a dataset will only suffer from one of these problems; it'll only suff
...
-A common problem is a dataset where some of the column names are not names of variables, but _values_ of a variable. Take `table4a`: the column names `1999` and `2000` represent values of the `year` variable, the values in the `1999` and `2000` columns represent values of the `cases` variable, and each row represents two observations, not one.
+A common problem is a dataset where some of the column names are not names of variables, but _values_ of a variable. Suppose you have your data in the following format.
 
 ```{r}
 table4a
 ```
 
+And you want to create the following visualisation where each line represents a `country`, `year` is on the x-axis, `cases` are on the y-axis, and you automatically get the legend that indicates which line represents which country.
+
+```{r tidy-pivot-longer-plot-lines, fig.width = 5, echo = FALSE}
```
hadley, line 135: For O'Reilly style, it's usually best to make floating figures with a caption.

tidy.Rmd
```diff
@@ -124,54 +124,120 @@ Typically a dataset will only suffer from one of these problems; it'll only suff
...
 Together those parameters generate the call to `pivot_longer()`:
 
 ```{r}
 table4a %>% 
-  pivot_longer(c(`1999`, `2000`), names_to = "year", values_to = "cases")
+  pivot_longer(
+    cols = c(`1999`, `2000`),
+    names_to = "year", 
+    values_to = "cases"
+    )
```
hadley, line 180: `)` should match indent of `pivot_wider()`

tidy.Rmd
```diff
@@ -124,54 +124,120 @@ Typically a dataset will only suffer from one of these problems; it'll only suff
...
 ```
 
-The columns to pivot are specified with `dplyr::select()` style notation. Here there are only two columns, so we list them individually. Note that "1999" and "2000" are non-syntactic names (because they don't start with a letter) so we have to surround them in backticks. To refresh your memory of the other ways to select columns, see [select](#select).
+The columns to pivot are specified with `dplyr::select()` style notation in the `cols` argument. Here there are only two columns, so we list them individually. Note that `1999` and `2000` are non-syntactic names (because they don't start with a letter) so we have to surround them in backticks. To refresh your memory of the other ways to select columns, see [select](#select).
```
hadley, line 183: Should probably change cross-links to `Section @\ref(#select)`.

tidy.Rmd
```diff
@@ -124,54 +124,120 @@ Typically a dataset will only suffer from one of these problems; it'll only suff
...
 
-```{r tidy-gather, echo = FALSE, out.width = "100%", fig.cap = "Pivoting `table4` into a longer, tidy form."}
+```{r tidy-pivot-longer, echo = FALSE, out.width = "100%", fig.cap = "Pivoting `table4a` into a \"longer\", tidy form."}
 knitr::include_graphics("images/tidy-9.png")
 ```
 
-In the final result, the pivoted columns are dropped, and we get new `year` and `cases` columns. Otherwise, the relationships between the original variables are preserved. Visually, this is shown in Figure \@ref(fig:tidy-gather). 
+There is still one issue though. Take a peek at the type of the `year` variable. We would expect `year` to be numeric (or specifically, we would expect it to be an integer), however it's showing up as a character. This is because the values in the `year` variable came from column headings in `table4a`. We can transform the type of this variable as we're pivoting the data using the `names_transform` argument in `pivot_longer()`. In this case we don't also need to transform the type of the `cases` variable, but if we did, we could simply add that to the list in `names_transform` as well.
```
hadley, line 193: Do you think it would be better to just use `mutate()` here? And call back to https://r4ds.had.co.nz/data-import.html#parsing-a-vector for other functions?

tidy.Rmd
mine-cetinkaya-rundel: Fixed throughout the chapter

tidy.Rmd
mine-cetinkaya-rundel: I think showing the `mutate()` option first is better, and then we could show you can achieve the same with `names_transform()`. Keeps it simple for those who want to stick to what they know but also points to additional `pivot_*()` functionality for the more adventurous reader. Does that sound good?

hadley: I'm not sure I like `names_transform` enough to even mention it here.

mine-cetinkaya-rundel: I am all for keeping it to just using `mutate()`. I think that feels like a more natural way to build a pipeline. 

I would suggest the same change further down in the case study section too then, just taking out `names_transform()` altogether and accomplishing its purpose with `pivot_*() %>% mutate()`.

hadley: Yeah, agreed.

mine-cetinkaya-rundel: Alright, all these changes are implemented now.

index.Rmd
```diff
@@ -14,7 +14,6 @@ documentclass: book
...
-
 This website is (and will always be) __free to use__, and is licensed under the [Creative Commons Attribution-NonCommercial-NoDerivs 3.0](http://creativecommons.org/licenses/by-nc-nd/3.0/us/) License.  If you'd like a __physical copy__ of the book, you can order it from [amazon](http://amzn.to/2aHLAQ1); it was published by O'Reilly in January 2017. If you'd like to __give back__
```
hadley, line 17: Was this change deliberate?

tidy.Rmd
```diff
@@ -62,27 +62,27 @@ Why ensure that your data is tidy? There are two main advantages:
...
+    Sections \@ref(mutate-funs) and \@ref(summarise-funs), most 
     built-in R functions work with vectors of values. That makes transforming 
     tidy data feel particularly natural.
 
 dplyr, ggplot2, and all the other packages in the tidyverse are designed to work with tidy data. Here are a couple of small examples showing how you might work with `table1`.
 
-```{r, out.width = "50%"}
+```{r fig.width = 5, fig.alt = "This figure shows the numbers of cases in 1999 and 2000 for Afghanistan, Brazil, and China, with year on the x-axis and number of cases on the y-axis. Each point on the plot represents the number of cases in a given country in a given year. The points for each country are differentiated from others by color and shape and connected with a line, resulting in three, non-parallel, non-intersecting lines. The numbers of cases in China are highest for both 1999 and 2000, with values above 200,000 for both years. The number of cases in Brazil is approximately 40,000 in 1999 and approximately 75,000 in 2000. The numbers of cases in Afghanistan are lowest for both 1999 and 2000, with values that appear to be very close to 0 on this scale."}
```
hadley, line 72: These are really nice illustrations of why you want both `fig.cap` and `fig.alt` 😄

index.Rmd
mine-cetinkaya-rundel: No, not deliberate. I'm not sure how that got in there when I did `pr_merge_main()`. Just pushed a fix to correct it.

hadley:
Thanks! Should appear momentarily on http://r4ds.hadley.nz

PR #918, opened 2021-01-26
Fix fct_lump example

mitsuoxv:
As n = 10 in fct_lump, the result should have 11 rows, but has only 10 rows, because `relig` happens to have "Other" in levels.


hadley:
I'd rather change the example to avoid this discussion altogether — I think it's a relatively complex issue that would be better left until later.

hadley:
OR actually, make it a footnote?

mitsuoxv:
How about an exercise?

hadley:
Great idea!

PR #919, opened 2021-02-02
.x, not ., is recommended in ~ formula

mitsuoxv:
.x is now recommended in ~ formula to avoid confusion with . in pipe


hadley:
Lovely, thanks!

PR #921, opened 2021-02-05
Fix inconsistency in season

mitsuoxv:
I think "summer" and "the Fall" are inconsistent, though I am not sure about articles.


hadley:
Thanks!

Issue #827, opened 2020-02-11
unnest() no longer fails when attempting to unnest list-columns of different lengths


mpaulacaldas:
The [last example](https://github.com/hadley/r4ds/blob/master/model-many.Rmd#L545-L552) in the "Many models" chapter no longer fails because `unnest()` recycles the shorter elements from the list-columns:

```r
# Doesn't work because y and z have different number of elements
df2 <- tribble(
  ~x, ~y,           ~z,
   1, "a",         1:2,  
   2, c("b", "c"),   3
)
df2
#> # A tibble: 2 x 3
#>       x y         z        
#>   <dbl> <list>    <list>   
#> 1     1 <chr [1]> <int [2]>
#> 2     2 <chr [2]> <dbl [1]>
df2 %>% unnest(y, z)
#> Warning: unnest() has a new interface. See ?unnest for details.
#> Try `df %>% unnest(c(y, z))`, with `mutate()` if needed
#> # A tibble: 4 x 3
#>       x y         z
#>   <dbl> <chr> <dbl>
#> 1     1 a         1
#> 2     1 a         2
#> 3     2 b         3
#> 4     2 c         3
```
I am not sure if this was an intentional change to the behaviour of `nest()` or if it's a bug that should be reported to tidyr. If the first, then the last section in the "Many models" chapter should be adapted to reflect the new behaviour of the function.

FixTestRepeat:
Hi, I found this ticket for similar reasons. Looks like the example for unnesting doesn't match the explanatory text anymore?

hadley:
This chapter is gone from the 2nd edition.

Issue #826, opened 2020-02-11
Adapt example relying on .drop argument of unnest()


mpaulacaldas:
The `.drop` argument of `unnest()` is deprecated, which raises a warning in the "Many models" chapter:

```r
glance <- by_country %>% 
  mutate(glance = map(model, broom::glance)) %>% 
  unnest(glance, .drop = TRUE)
#> Warning: The `.drop` argument of `unnest()` is deprecated as of tidyr 1.0.0.
#> All list-columns are now preserved.
#> This warning is displayed once per session.
#> Call `lifecycle::last_warnings()` to see where this warning was generated.
glance
#> # A tibble: 142 x 16
#> # Groups:   country, continent [710]
#>   country continent     data model resids r.squared adj.r.squared sigma
#>   <fct>   <fct>     <list<d> <lis> <list>     <dbl>         <dbl> <dbl>
#> 1 Afghan… Asia      [12 × 4] <lm>  <tibb…     0.948         0.942 1.22 
#> 2 Albania Europe    [12 × 4] <lm>  <tibb…     0.911         0.902 1.98 
#> 3 Algeria Africa    [12 × 4] <lm>  <tibb…     0.985         0.984 1.32 
#> 4 Angola  Africa    [12 × 4] <lm>  <tibb…     0.888         0.877 1.41 
#> 5 Argent… Americas  [12 × 4] <lm>  <tibb…     0.996         0.995 0.292
#> 6 Austra… Oceania   [12 × 4] <lm>  <tibb…     0.980         0.978 0.621
#> # … with 136 more rows, and 8 more variables: statistic <dbl>, p.value <dbl>,
#> #   df <int>, logLik <dbl>, AIC <dbl>, BIC <dbl>, deviance <dbl>,
#> #   df.residual <int>
```
The line is:

https://github.com/hadley/r4ds/blob/9dda73b31f7ff0a8d3f1ab3ffe3bdb1110a4ca2a/model-many.Rmd#L220

This also affects [this exercise](https://github.com/hadley/r4ds/blob/master/model-many.Rmd#L266-L270).

Since the documentation of `nest()` suggests relying on `select()` instead `.drop = TRUE`, I think the example should either be removed or adapted to prevent confusing readers.

Of course, I would be more than happy to propose a PR depending on your decision!

hadley:
This chapter is now gone 😄

PR #922, opened 2021-02-06
Fix unnest examples

mitsuoxv:
unnest: .drop no longer works, multiple columns require not equal but compatible lengths, and some typos


mitsuoxv:
Fix #826, fix #827

mine-cetinkaya-rundel:
The modeling chapters will be out of scope for the next edition of the book so I'll close this PR, but thank you for the contribution!

PR #927, opened 2021-02-27
Fixed Typo

mayankvanani:
Fixed a minor typo.


mine-cetinkaya-rundel:
Actually, this is intentional, to make it clear that you can indeed manage without the spaces, but it makes it clear to have them in place. Thanks anyway though!

PR #937, opened 2021-03-26
Fix str_to_upper’s default locale

sonicdoe:
According to [stringr’s reference](https://stringr.tidyverse.org/reference/case.html), the locale defaults to English:

> **locale**: Locale to use for translations. Defaults to "en" (English) to ensure consistent default ordering across platforms.


mine-cetinkaya-rundel:
Thank you!

PR #943, opened 2021-04-14
Remove "dumb" from acknowledgements.

camillevleonard:
I was just reading through this section of the book and thought that it would be good to remove "dumb". I'm trying to remove ableist language from my every day vocabulary and thought the R team would be interested in that as well.


mine-cetinkaya-rundel:
Thank you!

PR #952, opened 2021-05-15
Change "with a" to "that have" for clarity

gabrivera:
In the section 3.6 on Geometrical objects, I momentarily understood that it was saying that one line is describing the entirety of the values in the data-set through the 4 value. This is unlikely, due to the context, but it happened to me before realizing that wasn't the case. This small change could avoid that misunderstanding.


mine-cetinkaya-rundel:
Thanks!

PR #958, opened 2021-06-04
Defined "word" in exercise 14.4.4.1.3

ReeceGoding:
Changed "Which of those strings are still words?" to "Which of those strings are still `words`?". This gives a big hint to the reader, hopefully letting them figure out how to solve this part of the exercise in an automatic way, rather than trying to eyeball an answer.


hadley:
Thanks!

PR #961, opened 2021-08-30
Change ", but" to "and" in intro.Rmd

EricKit:
First time reading this book - thank you!

The `,but` did not read well for me, as the statement 

>  [visualisations] don't scale particularly well because they require a human to interpret them

does not appear to conflict with the statement

> Visualisations can surprise you

`and` appeared more appropriate as these statements seem to be two outcomes from visualizations; unless I am wrong about the intent of the sentence.

An alternative version, which I did not think was the intent, would read

> Visualisations can surprise you, but don't scale particularly well, because they require a human to interpret them.


mine-cetinkaya-rundel:
Thank you!

PR #964, opened 2021-09-17
consistent with treatment of *alpha* ?

mbeveridge:
add emphasis, to be consistent with treatment of *alpha* ? ...

row189 : "or to the shape aesthetic" --> "or to the *shape* aesthetic"

[sorry if I've forgotten how to use github]


mine-cetinkaya-rundel:
Thank you!

PR #973, opened 2021-11-18
Added link ggplot2 cheat sheet mention

GeroVanMi:
Added the link to http://rstudio.com/resources/cheatsheets to the mention under "3.7 Statistical transformations" in order for people to be able to jump to the correct source directly, instead of having to navigating back up.

Didn't add it as an inline link in consideration of offline (book) readers.


hadley:
Inline links will be automatically converted to footnotes in the book, so an inline link would be better here.

GeroVanMi:
> Inline links will be automatically converted to footnotes in the book, so an inline link would be better here.

@hadley 
That's even better, thanks for the feedback.  
I've changed it to an inline-link.

hadley:
Thanks!

PR #980, opened 2021-12-04
Updates order information

medbox1:
Changes To be published/pre-order language to order language.


mine-cetinkaya-rundel:
Thank you, we've already updated the relevant bit, so I'll close this PR.

PR #982, opened 2021-12-15
Update EDA.Rmd

PursuitOfDataScience:
All three Amazon book links are outdated. I fixed all of them.


mine-cetinkaya-rundel:
Fixed some of these in 323ce2175b0987686b9ef1d80ae638ec8622542c with slightly simplified links and linking to online versions where available. Thank you for the pointers!

PR #984, opened 2021-12-20
Update program.Rmd

PursuitOfDataScience:
Book links do not work as of now. I replaced the links that can direct viewers to the right sites!


mine-cetinkaya-rundel:
Thank you!

PR #987, opened 2021-12-23
Update EDA.Rmd

PursuitOfDataScience:
The ggplot2 book is freely accessible. Therefore, some text in this document isn't applicable.


hadley:
Fixed in another PR; thanks for working on this and sorry I didn't merge it earlier.

PR #998, opened 2022-03-08
Few edits that may be helpful

mosesotieno:
These are few edits that may be helpful


hadley:
These are all in the dev version of stringr

PR #1006, opened 2022-03-29
Remove obsolete sentence

sonicdoe:
[ggplot 2.2.0 has been available since 2016](https://www.rstudio.com/blog/ggplot2-2-2-0/).


hadley:
Good catch, thanks!

PR #1008, opened 2022-04-07
Replace "data frame" with "`data.frame`" in the Tibbles chapter

Iain-S:
In the opening of the Tibbles chapter, it is said

> when I want to draw particular attention to R’s built-in data frame, I’ll call them data.frames

so it could aid clarity to do that for the remainder of the page.


hadley:
Thanks!

PR #1012, opened 2022-04-27
Minor clarification in chapter 14.

c-hoh:
The text talks about delayed flights, while the data and the plot is about cancelled flights.


hadley:
Thanks!

PR #1020, opened 2022-05-07
Update preface-2e.Rmd

AnttiRask:
Correction for the line about new chapters on column-wise and row-wise operations.


mine-cetinkaya-rundel:
Thank you!

PR #1028, opened 2022-05-13
Exchange `ahead` and `behind`, and typos in Logical vectors

mitsuoxv:
Isn't it natural to name the average delay time of actually delayed flights as `behind', and that of early arrival flights as `ahead`?


mine-cetinkaya-rundel:
Thank you! I agree that the ahead/behind naming seemed to be the opposite of intended. I've implemented those updates and other typo fixes in the associated .qmd file.

PR #1031, opened 2022-05-13
Use %% as stated, recycling rules, and typos in numbers.Rmd

mitsuoxv:
15.3.2 Rank; last sentence: "When combined with %% or %/%, ..." So I use %% in the following chunk.
15.4.6 mutate(); As recycling rules are discussed in this chapter, I drop "we discussed in Section 30.4.3". Or it may be better to move Chapter 30 Vectors to Transform.
15.4.5 Positions; Although R ] (subset) function is already used in Chapter 14 logical vectors, I keep "we'll not cover it until Section 30.4.5". This may also be solved by moving Chapter 30 Vectors to Transform.

### DIFF
--- a/numbers.Rmd
+++ b/numbers.Rmd
@@ -110,7 +110,7 @@ There are a couple of variants of `n()` that you might find useful:
 
 Transformation functions work well with `mutate()` because their output is the same length as the input.
 The vast majority of transformation functions are already built into base R.
-It's impractical to list them all so this section will give show the most useful.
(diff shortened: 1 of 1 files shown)


mine-cetinkaya-rundel:
Implemented the changes in the associated .qmd, thank you!

wszzcdsg:
这是来自QQ邮箱的假期自动回复邮件。你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复。

PR #1033, opened 2022-05-26
Minor typos

mine-cetinkaya-rundel:
Only minor typos fixed in PR.

One note: There is a sentence that says "For example, take this pipeline that filters on a summary value." -- that should be followed by something? (Right before joins).


wszzcdsg:
这是来自QQ邮箱的假期自动回复邮件。你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复。

PR #1043, opened 2022-06-13
Correct function name

waltersom:
Correct function names `str_lower` and `to_lower` to `str_to_lower`


mine-cetinkaya-rundel:
Thank you!

PR #1048, opened 2022-06-21
Update data-visualize.qmd

### LINKED ISSUE #1044: Minor edit in data-visualize.qmd
Line 91 in data-visualize.qmd describes a positive relationship, when it would be clearer for it to describe a negative relationship, as specified in the previous line: https://github.com/hadley/r4ds/blob/main/data-visualize.qmd#L90-L91.

I would propose deleting line 91 and replacing it with "In other words, cars with small engines have higher fuel efficiency and, in general, as engine size increases, fuel efficiency decreases."

If that change sounds reasonable, I'll gladly open up a pull request for the change.

kevinxperese:
Describe a negative relationship between engine size and fuel efficiency. Fixes #1044.

### DIFF
--- a/data-visualize.qmd
+++ b/data-visualize.qmd
@@ -88,7 +88,7 @@ ggplot(data = mpg) +
 ```
 
 The plot shows a negative relationship between engine size (`displ`) and fuel efficiency (`hwy`).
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1063, opened 2022-08-15
Update regexps.qmd

PursuitOfDataScience:
Removed repetitious part.


hadley:
Thanks for all these little fixes!

PR #1065, opened 2022-08-15
Update regexps.qmd

PursuitOfDataScience:
Fixed a few errors. Could you also explain why there are two dots in "..\\1" rather than three dots or some other numbers of dots?


hadley:
Just because I'm looking for repeating **pairs** of letters.

PursuitOfDataScience:
Yes, I figured that out right after I made comment. Thanks!

PR #1070, opened 2022-08-16
Typo in parameter name

dominicwhite:
Rename `dota` to `data` in `ggplot()`


hadley:
The point of that exercise is to figure it out 😄

PR #1071, opened 2022-08-16
Update iteration.qmd

PursuitOfDataScience:
Provided another way of getting R squared by using `nest()` from tidyr, which makes the code to be tidyverse consistent.


hadley:
Thanks! (Note this chapter has yet to be touched for the 2e, so it might get more substantial changes)

PR #1075, opened 2022-08-18
Update data-transform.qmd - Added installation of package code

quinten-goens:
Upon reading the text I was (as a beginner) a bit surprised that my code did not work. I first needed to install the library which was not clear to me, hence this change.


quinten-goens:
Noted that this was in the intro.

PR #1080, opened 2022-08-28
Update functions.qmd

august-18:
"As you write more and more functions you’ll eventually want to convert these informal, interactive tests into formal, automated tests. That process is called unit testing. Unfortunately, it’s beyond the scope of this book, but you can learn about it in http://r-pkgs.had.co.nz/tests.html."
Link gives 404 error for page not found
I updated to https://r-pkgs.org/testing-basics.html


mine-cetinkaya-rundel:
Thank you!

PR #1084, opened 2022-09-03
Fix typos

betulturkoglu:
I fixed inline R codes and file type inconsistency in sentences at [https://r4ds.hadley.nz/rmarkdown-formats.html#documents](https://r4ds.hadley.nz/rmarkdown-formats.html#documents).


hadley:
Looks like we fixed most of these in the meantime, so I merged the one last fix by hand. Thanks for contributing!

betulturkoglu:
You are welcome!

PR #1089, opened 2022-09-14
Fix pipe terminology in plaintext

zekiakyol:
Co-Authored-By: Betul Turkoglu <79486373+betulturkoglu@users.noreply.github.com>

We proofread the book for the third time and
- Since `|>` is referred as "native pipe" in [RStudio Options](https://r4ds.hadley.nz/workflow-pipes.html#fig-pipe-options) and is also explained that it is a part of base R in the book, we think it would be better to say `"native pipe"` instead of `"base pipe"` in plaintext to keep the terminology consistent.
- Fixed plaintext typos and inline R codes.

### DIFF
--- a/data-tidy.qmd
+++ b/data-tidy.qmd
@@ -199,7 +199,7 @@ billboard |>
 ```
 
 What happens if a song is in the top 100 for less than 76 weeks?
-Take 2 Pac's "Baby Don't Cry", for example.
+Take 2Pac's "Baby Don't Cry", for example.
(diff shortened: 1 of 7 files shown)


hadley:
Hmmmm, we use the term base pipe everywhere else, so it might be better to leave this inconsistency with RStudio in.

zekiakyol:
Thank you for the feedback! We will adopt `"base pipe"` from now on.
Since this pull request failed render check, I close it.

PR #1091, opened 2022-09-15
typos in iteration.qmd

martin-raden:


hadley:
It's not worth proofing this chapter as I'm currently heavily restructuring it.

PR #1093, opened 2022-09-17
Update iteration.qmd

PursuitOfDataScience:
Made a few edits.


hadley:
Thanks for providing these fixes, but it's not worth tweaking these chapters that I'm currently heavily rewriting. It unfortunately tends to lead to merge commits because I've make big changes locally.

PR #1096, opened 2022-09-21
fig_alt fix: color -> size

jooyoungseo:
I could be wrong, but the legend description seems incorrect because different size corresponds to levels of `class` variable. Is color mapping also used in this legend?


hadley:
Ooops, thanks!

PR #1101, opened 2022-10-11
Made some grammar changes

cskaust:
some sentences in the comments needed more proper english


hadley:
Thanks — the code of conduct comes from https://www.contributor-covenant.org, so I left as is.

PR #1103, opened 2022-10-19
Mention arrow in "Big data" section

andreranza:
Maybe arrow *for R* would be better?


hadley:
Thanks for suggesting this — we're going to give arrow it's own chapter.

PR #1106, opened 2022-10-21
fixing typos in functions.qmd

mnazarov:
In addition to these, there is also inconsistent use of `summarise` and `summarize`, but maybe this is on purpose to show that there are 2 spelling options?


hadley:
Thanks — and it's definitely not on purpose.

PR #1113, opened 2022-10-28
New chapter on important base R functions

hadley:


base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+
+We'll focus on three big topics: subsetting with `[`, subsetting with `[[` and `$` and iteration with for loops and the apply family of functions.
+These tools are particularly useful as you being to do more programming in R, which is why we don't teach them earlier.
+
+This is a good place to remind you that the tidyverse is not only way to solve data science problems.
+We teach the tidyverse in this book because it's designed as whole, which tends to increase consistency across functions making it a little easier to use.
+After you read this book we expect you'll learn other ways to tackle the same problems using base R, data.table, and other packages.
+You'll certainly encounter these other approaches when you start reading R code written by other people, particularly if you're using StackOverflow.
+It's 100% OK for your code to include a mix of paradigms, but when you have time, we highly recommend taking the time to rewrite with a consistent style --- this will help you understand what it does better, improve your R programming skills, and lead to code that's easier to maintain.
```
DavisVaughan, line 20: I feel like I wanted to see something here about the fact that the tidyverse would not be possible without base R.

Call it what you want, but the tidyverse is built on top of base R code (if you count the R C API as base R code, which I do).

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+
+df2 <- tibble(x = 1:3)
+df2[, "x"]
+```
+
+You can also avoid this problem by using 1d `[` to select columns:
+
+```{r}
+df1["x"]
+```
```
DavisVaughan, line 140: I guess this is easier to explain than `drop = FALSE`?

DavisVaughan: May be worth a sentence explicitly acknowledging the difference between `x[cols]` and `x[rows,]`

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+
+## Selecting a single element `$` and `[[`
+
+`[`, which selects many elements, is paired with `[[` and `$`, which extract a single element from a list or data frame.
+In this section, we'll show you how to use `[[` and `$` to pull columns out of a data frames, a couple of differences between `data.frames` and tibbles, and how `[` differs from `[[` when used with lists.
+
+### Data frames
+
+`[[` and `$` can be used like `pull()` to extract columns out of a data frame.
+`[[` can access by position or by name, and `$` is specialized for access by position:
```
DavisVaughan, line 213:
```suggestion
`[[` can access by position or by name, and `$` is specialized for access by name:
```

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+# Then transform each column with lapply() and replace the original values
+df[num_cols] <- lapply(df[num_cols], \(x) x * 2)
+df
+```
+
+The code above uses a new function, `sapply()`.
+It's similar to `lapply()` but it always tries to simplify the result, hence the `s` in its name, here producing a logical vector instead of a list.
+We don't recommend using it for programming, because the simplification can fail and give you an unexpected type, but it's usually fine for interactive use.
+
+Base R provides a stricter version of `sapply()` called `vapply()`, short for vector apply.
```
DavisVaughan, line 392: Could draw some connections between `sapply->map_vec` and `vapply->map_int/dbl` if you wanted to

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+diamonds |> 
+  group_by(cut) |> 
+  summarise(price = mean(price))
+
+tapply(diamonds$price, diamonds$cut, mean)
+```
+
+Unfortunately it returns its results in a named vector which requires some gymnastics if you want to collect multiple summaries into a data frame matched with the variable you're grouping by.
+(It's certainly possible to avoid this and just work with free floating vectors, but that generally just makes more work down the road).
+If you want to see how you might use `tapply()` or other base techniques to perform other grouped summaries, Hadley has collected a few techniques [in a gist](https://gist.github.com/hadley/c430501804349d382ce90754936ab8ec).
```
DavisVaughan, line 412: My favorite `tapply()` feature is that when given factors to "group" by, you get one summary result per factor level, even if that factor level didn't appear in the actual data. Which made it useful in hardhat for `weighted_table()`
https://github.com/tidymodels/hardhat/blob/1e6774ce9b5db35ce36a6247cd9390cd23db75da/R/table.R#L126

Somewhat like `dplyr::group_by(.drop = FALSE)`

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+
+We'll focus on three big topics: subsetting with `[`, subsetting with `[[` and `$` and iteration with for loops and the apply family of functions.
+These tools are particularly useful as you being to do more programming in R, which is why we don't teach them earlier.
+
+This is a good place to remind you that the tidyverse is not only way to solve data science problems.
+We teach the tidyverse in this book because it's designed as whole, which tends to increase consistency across functions making it a little easier to use.
+After you read this book we expect you'll learn other ways to tackle the same problems using base R, data.table, and other packages.
+You'll certainly encounter these other approaches when you start reading R code written by other people, particularly if you're using StackOverflow.
+It's 100% OK for your code to include a mix of paradigms, but when you have time, we highly recommend taking the time to rewrite with a consistent style --- this will help you understand what it does better, improve your R programming skills, and lead to code that's easier to maintain.
+
```
DavisVaughan, line 21: Maybe some reminder that you've really been using base R all along throughout this book. The entire `Transform` section uses data types and functions that come from base R, i.e. we didn't invent factors/Date objects, and we didn't invent the `&` or `>` functions.

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+tb$x
+tb$z
+```
+
+For this reason we sometimes joke that tibbles are lazy and surly: they do less and complain more.
+
+### Lists
+
+`[[` and `$` are also really important for working with lists, and it's really important to understand how they differ to `[`.
+Lets illustrate the differences with a list named l:
```
mine-cetinkaya-rundel, line 264:
```suggestion
Let's illustrate the differences with a list named `l`:
```

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+
+```{r}
+l <- list(a = 1:3, b = "a string", c = pi, d = list(-1, -5))
```
mine-cetinkaya-rundel, line 267:
```suggestion
l <- list(
  a = 1:3, 
  b = "a string", 
  c = pi, 
  d = list(-1, -5)
)
```

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+
+    ```{r}
+    str(l[1:2])
+    str(l[4])
+    ```
+
+    Like with vectors, you can subset with a logical, integer, or character vector.
+
+-   `[[` and `$` extract a single component from a list.
+    They removes a level of hierarchy from the list.
```
mine-cetinkaya-rundel, line 281:
```suggestion
    They remove a level of hierarchy from the list.
```

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+    str(l[[1]])
+    str(l[[4]])
+
+    str(l$a)
+    ```
+
+The difference between `[` and `[[` is important for lists because `[[` drills down into the list while `[` returns a new, smaller list.
+This difference very important, but it's easy to get confused.
+To help you remember, take a look at the an unusual pepper shaker shown in @fig-pepper-1.
+If this pepper shaker is your list `pepper`, then, `pepper[1]` is a pepper shaker containing a single pepper packet, as in @fig-pepper-2.
```
mine-cetinkaya-rundel, line 293:
```suggestion
If we suppose this pepper shaker is a list `pepper`, then, `pepper[1]` is a pepper shaker containing a single pepper packet, as in @fig-pepper-2.
```

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+They're wrong!
+(Well at least they're rather out of date, as `for` loops haven't been slow for many years.) The chief benefit of using functions like `map()` or `apply()` is not speed, but clarity: once you've mastered the basic idea, they make your code faster to write and easier to read.
+
+### Apply family
+
+Another way of saying "map a function over each element of a vector" is "apply a function over each element of a vector", and base R provides a family of functions known as the "apply" family because they all use apply as a suffix.
+Here we'll give you a quick overview of this family so you can recognize them in the wild.
+
+The most important member of this family is `lapply()`, which is very similar to `purrr::map()`[^base-r-3].
+In fact, because we don't use any of `map()`'s more advanced features, you can replace every `map()` call in @sec-iteration with `lapply()`.
```
mine-cetinkaya-rundel, line 370:
```suggestion
In fact, because we haven't used any of `map()`'s more advanced features, you can replace every `map()` call in @sec-iteration with `lapply()`.
```

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+df
+```
+
+The code above uses a new function, `sapply()`.
+It's similar to `lapply()` but it always tries to simplify the result, hence the `s` in its name, here producing a logical vector instead of a list.
+We don't recommend using it for programming, because the simplification can fail and give you an unexpected type, but it's usually fine for interactive use.
+
+Base R provides a stricter version of `sapply()` called `vapply()`, short for vector apply.
+It takes an additional argument that specifies the expected type which ensures that simplification occurs the same way regardless of the input.
+For example, we could replace the `sapply()` call above with this `vapply()` where we specific that we expect `is.numeric()` to return a logical vector of length 1:
```
mine-cetinkaya-rundel, line 394:
```suggestion
For example, we could replace the `sapply()` call above with this `vapply()` where we specificy that we expect `is.numeric()` to return a logical vector of length 1:
```

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+## Summary
+
+<https://adv-r.hadley.nz/subsetting.html>
+
+<https://adv-r.hadley.nz/control-flow.html#loops>
+
+This chapter concludes the programming section of the book.
+You've now learned the basics of programming in R.
+You know now the data types that underpin all of the objects you work with, and have two powerful techniques (functions and iteration) for reducing the duplication in your code.
+We hope you've got a taste for how programming can help your analyses, and you've made a solid start on your journey to become not just a data scientist who uses R, but a data science who can program in R.
```
mine-cetinkaya-rundel, line 502:
```suggestion
We hope you've got a taste for how programming can help your analyses, and you've made a solid start on your journey to becoming not just a data scientist who uses R, but a data scientist who can program in R.
```

base-R.qmd
```diff
@@ -0,0 +1,502 @@
...
+
+1.  Create functions that take a vector as input and return:
+
+    a.  The elements at even numbered positions.
+    b.  Every element except the last value.
+    c.  Only even numbers (and no missing values).
+
+2.  Why is `x[-which(x > 0)]` not the same as `x[x <= 0]`?
+    Read the documentation for `which()` and do some experiments to figure it out.
+
```
mine-cetinkaya-rundel, line 204: Additional exercise idea:

Suppose you have the following data frame:

```r
df <- tibble(x = c(1, 2, NA, 3, 4, 5, NA))
```

Subset for rows where `x > 2` using `filter()` and also using `[`. How do the results compare?

mine-cetinkaya-rundel (approved):
Looks great. Made a bunch of minor edits, might be easier to accept here and pull. Also left a suggestion for an exercise.

Overall I think this chapter does cover the important stuff from base R. The one other thing we might cover is `plot()` or `hist()` -- something for a very simple plot with some commentary like "you might run into them in the wild".

base-R.qmd
```diff
@@ -0,0 +1,536 @@
...
+These tools are particularly useful as you do more programming and will help you read code that you'll encounter in the wild.
+
+This is a good place to remind you that the tidyverse is not only way to solve data science problems.
+We teach the tidyverse in this book because tidyverse packages share a common design philosophy, which increases the consistency across functions, making each new function or package a little easier to learn and use.
+It's not possible to use the tidyverse without using base R, so we've already taught you a **lot** of base R functions: from `library()` to load packages, to `sum()` and `mean()` for numeric summaries, to the factor, date, and POSIXct data types, and of course all the basic operators like `+`, `-`, `/`, `*`, `|`, `&`, and `!`.
+
+After you read this book you'll learn other approaches to the same problems using base R, data.table, and other packages.
+You'll certainly encounter these other approaches when you start reading R code written by other people, particularly if you're using StackOverflow.
+It's 100% okay to write code that uses a mixes of approaches, and don't let anyone tell you otherwise!
```
mine-cetinkaya-rundel, line 19:
```suggestion
It's 100% okay to write code that uses a mix of approaches, and don't let anyone tell you otherwise!
```

base-R.qmd
```diff
@@ -0,0 +1,536 @@
...
+source("_common.R")
+status("polishing")
+```
+
+To finish off the programming section, we're going to give you a quick tour of the most important base R functions that we don't otherwise discuss in the book.
+These tools are particularly useful as you do more programming and will help you read code that you'll encounter in the wild.
+
+This is a good place to remind you that the tidyverse is not only way to solve data science problems.
+We teach the tidyverse in this book because tidyverse packages share a common design philosophy, which increases the consistency across functions, making each new function or package a little easier to learn and use.
+It's not possible to use the tidyverse without using base R, so we've already taught you a **lot** of base R functions: from `library()` to load packages, to `sum()` and `mean()` for numeric summaries, to the factor, date, and POSIXct data types, and of course all the basic operators like `+`, `-`, `/`, `*`, `|`, `&`, and `!`.
```
mine-cetinkaya-rundel, line 15:
```suggestion
It's not possible to use the tidyverse without using base R, so we've actually already taught you a **lot** of base R functions: from `library()` to load packages, to `sum()` and `mean()` for numeric summaries, to the factor, date, and POSIXct data types, and of course all the basic operators like `+`, `-`, `/`, `*`, `|`, `&`, and `!`.
What we haven't focused on so far is base R workflows, so we will highlight a few of those in this chapter.
```

mine-cetinkaya-rundel (approved):
Re-review intro

PR #1123, opened 2022-11-07
base-R typos and comments

jonthegeek:
Still in progress but saving before I lose my place (I was editing in the web interface while I read).

Line 108: I think this is the first use of `runif()` in the book (from a quick search). Either use a different example, or explain it. Maybe you can catch them before they read it as "run if" and get confused by it for a few months!

Line 126: I don't understand what you're trying to say here:
> so when we want to draw particular attention to R's built-in data frame, we'll write `data.frame`s.

Do you mean that you'll call the `data.frame()` function during construction?


hadley:
I'm just trying to point out our notation convention for `data.frame`s.

hadley:
Do you want to finish this off? 😄

jonthegeek:
Doh, sorry! I'll wrap this up tomorrow!

jonthegeek:
I'd love it if you could provide a preferred & memorable pronunciation of `[`, `[[`, and `$` (in context; what do you think when you read `x[1]` vs `x[[1]]` vs `x$1`?). I've been trying to take Felienne's Hermans's advice to learn to pronounce code to make it easier to think in code, but I've never found anything for those that's memorable and meaningful.

In the for loop section, you use "for loop", "`for()` loop", "for-loop", and "`for` loop" at various points. I don't *think* they had distinct meanings, so probably standardize on one of them.

Great chapter overall! Thanks for adding it!

hadley:
Thanks!

Unfortunately I don't have a pronunciation that I use in my head so I can't suggest anything here 😞

PR #1124, opened 2022-11-07
Change "locations" to "indexes"

kwstat:
The word "locations" makes it sound like `str_locate()`, which is NOT the case.


hadley:
Thanks!

PR #1127, opened 2022-11-08
Update strings.qmd

jeromecholewa:
Sorry, these updates are only cosmetic , not on the code. :)

str_glue instead of glue (mistype?)
delimiter instead of delimater
combining instead of combing


hadley:
Thank you!

PR #1128, opened 2022-11-08
Eliminate parsing chapter

hadley:
Originally the plan was to have two chapters about reading text files, a brief introduction in the whole game, and then a more detailed exploration later in the book. This organisation didn't seem to work very well because the second chapter didn't have much content, so I've removed it, integrating its content elsewhere in the book:

* Column parsing types moved back into data-import
* Specifics of parsing various data types (e.g. `col_number()`, `col_date()`, and `col_factor()`) moved into the corresponding data type chapters.
* String encoding has moved to the strings chapter

While I was in here I also removed the unused `import-other.qmd`; we had planned to survey other options but I no longer think this is worth it.


jennybc (commented):
As directed, I went very fast and mostly just made sure everything about readr sounds right.

You might want to add is `cols_only()`, where you show switching to `cols()` in order to change the default type.

Could also mention the general practice of letting guessing happen, getting that spec, maybe editing it?, then plugging it back into your code for greater type safety in future.

data-import.qmd
```diff
@@ -64,165 +48,165 @@ read_csv("data/students.csv") |>
...
+)
 ```
 
-Note that the values in the `meal_type` variable has stayed exactly the same, but the type of variable denoted underneath the variable name has changed from character (`<chr>`) to factor (`<fct>`).
-
-Before you move on to analyzing these data, you'll probably want to fix the `age` column as well: currently it's a character variable because of the one observation that is typed out as `five` instead of a numeric `5`.
-We discuss the details of fixing this issue in @sec-import-spreadsheets in further detail.
-
-### Compared to base R
+These arguments are all you need to know to read \~80% of CSV files that you'll encounter in practice.
```
mine-cetinkaya-rundel, line 191: Since 80% is likely a made-up number, how about "a large majority of CSV files" instead?

data-import.qmd
```diff
@@ -269,6 +253,118 @@ df |> pivot_longer(everything())
     )
     ```
 
+## Controlling column types {#sec-col-types}
+
+A csv file doesn't contain any information about the type of each variable (i.e. whether it's a logical, number, string etc), so readr will try to guess the type.
```
mine-cetinkaya-rundel, line 258:
```suggestion
A CSV file doesn't contain any information about the type of each variable (i.e. whether it's a logical, number, string, etc), so readr will try to guess the type.
```

data-import.qmd
```diff
@@ -269,6 +253,118 @@ df |> pivot_longer(everything())
...
+```
+
+This heuristic works well if you have a clean dataset, but in real life you'll encounter a selection of weird and wonderful failures.
+
+### Missing values, column types, and problems
+
+The most common way column detection fails is that a column contains unexpected values and you get a character column instead of a more specific type.
+One of the most common causes for this a missing value, recorded using something other than the `NA` that stringr expects.
+
+Take this simple 1 column csv file as an example:
```
mine-cetinkaya-rundel, line 293:
```suggestion
Take this simple 1 column CSV file as an example:
```

data-import.qmd
```diff
@@ -269,6 +253,118 @@ df |> pivot_longer(everything())
...
+
+```
+
+If we read it without any additional arguments, `x` becomes a character column:
+
+```{r}
+df <- read_csv(csv)
+```
+
+In this very small case, you can easily see that the missing value `.`.
```
mine-cetinkaya-rundel, line 311:
```suggestion
In this very small case, you can easily see the missing value `.`.
```

data-import.qmd
```diff
@@ -269,6 +253,118 @@ df |> pivot_longer(everything())
...
+But what happens if you have thousands of rows with only a few speckled amongst them?
```
mine-cetinkaya-rundel, line 312:
```suggestion
But what happens if you have thousands of rows with only a few missing values represented by `.`s speckled amongst them?
```

data-import.qmd
```diff
@@ -269,6 +253,118 @@ df |> pivot_longer(everything())
...
+
+Now `read_csv()` reports that there was a problem, and tells us we can find out more with `problems()`:
+
+```{r}
+problems(df)
+```
+
+This tells us that there was a problem in row 3, col 1 where readr expected a double but got a `.`.
+That suggests this dataset uses `.` for missing values.
+So when we set `na = "."`, the automatic guessing succeeds, giving us the numeric column that we want:
```
mine-cetinkaya-rundel, line 328:
```suggestion
So then we set `na = "."`, the automatic guessing succeeds, giving us the numeric column that we want:
```

data-import.qmd
```diff
@@ -335,14 +425,14 @@ There are two alternatives:
     read_rds("students.rds")
     ```
 
-2.  The feather package implements a fast binary file format that can be shared across programming languages:
+2.  The arrow package allows you to read and write parquet files, a fast binary file format that can be shared across programming languages:
```
mine-cetinkaya-rundel, line 428: Should add cross-ref to the new binary data chapter?

datetimes.qmd
```diff
@@ -74,20 +74,84 @@ today()
 now()
 ```
 
-Otherwise, there are three ways you're likely to create a date/time:
+Otherwise, there following sections describe the four ways you're likely to create a date/time:
```
mine-cetinkaya-rundel, line 77:
```suggestion
Otherwise, the following sections describe the four ways you're likely to create a date/time:
```

datetimes.qmd
```diff
@@ -74,20 +74,84 @@ today()
...
 
+-   While reading a file with readr.
 -   From a string.
 -   From individual date-time components.
 -   From an existing date/time object.
 
-They work as follows.
+### During import
+
+If your csv contains an ISO8601 date or date-time, you don't need to do anything; readr will automatically recognize it:
```
mine-cetinkaya-rundel, line 86:
```suggestion
If your CSV contains an ISO8601 date or date-time, you don't need to do anything; readr will automatically recognize it:
```

numbers.qmd
```diff
@@ -30,6 +30,27 @@ library(tidyverse)
 library(nycflights13)
 ```
 
+## Making numbers
+
+In most cases, you'll get numbers already recorded in one R's numeric types, integer or double.
```
mine-cetinkaya-rundel, line 35:
```suggestion
In most cases, you'll get numbers already recorded in one of R's numeric types: integer or double.
```

numbers.qmd
```diff
@@ -30,6 +30,27 @@ library(tidyverse)
...
+In some cases, however, you'll encounter them as strings, possibly because you've created the by pivotting from column headers or something has gone wrong in your data import process.
```
mine-cetinkaya-rundel, line 36:
```suggestion
In some cases, however, you'll encounter them as strings, possibly because you've created them by pivoting from column headers or something has gone wrong in your data import process.
```

numbers.qmd
```diff
@@ -753,9 +774,9 @@ For example:
 
 ## Summary
 
-You're likely already familiar with many tools for working with numbers, and in this chapter you'll have learned how they're realized in R.
-You also learned a handful of useful general transformations that are commonly, but not exclusively, applied to numeric vector like ranks and offsets.
-Finally, we worked through a number of numeric summaries, and discussed a few of the statistical challenges that you should consider.
+You're already familiar with many tools for working with numbers, and after reading this chapter you now know how to use them in R.
+You've also learned a handful of useful general transformations that are commonly, but not exclusively, applied to numeric vector like ranks and offsets.
```
mine-cetinkaya-rundel, line 778:
```suggestion
You've also learned a handful of useful general transformations that are commonly, but not exclusively, applied to numeric vectors like ranks and offsets.
```

strings.qmd
```diff
@@ -14,7 +14,7 @@ Now it's time to dive into them, learning what makes strings tick, and mastering
 
 We'll begin with the details of creating strings and character vectors.
 You'll then dive into creating strings from data, then the opposite; extracting strings from data.
-The chapter finishes up with functions that work with individual letters and a brief discussion of where your expectations from English might steer you wrong when working with other languages.
+We'll then discuss tools that work with individual letters, and finish off with a brief discussion of how English The chapter finishes up with functions that work with individual letters and a brief discussion of where your expectations from English might steer you wrong when working with other languages.
```
mine-cetinkaya-rundel, line 17:
```suggestion
We'll then discuss tools that work with individual letters. The chapter finishes off with functions that work with individual letters and a brief discussion of where your expectations from English might steer you wrong when working with other languages.
```

strings.qmd
```diff
@@ -451,12 +451,8 @@ df |>
...
 ## Letters
 
-This section discusses stringr functions that work with individual letters.
-This is straightforward for English because it uses an alphabet with 26 letters, but things rapidly get complicated when you move beyond English.
-Even languages that use the same alphabet but add additional accents (e.g. å, é, ï, ô, ū) are non-trivial because those letters might be represented as an individual character or by combining an unaccented letter (e.g. e) with a diacritic mark (e.g. ´).
-And other languages "letters" look quite different: in Japanese each "letter" is a syllable, in Chinese each "letter" is a complex logogram, and in Arabic letters look radically different depending on their location in the word.
-
-In this section, we'll assume that you're working with English text as we introduce to functions for finding the length of a string, extracting substrings, and handling long strings in plots and tables.
+In this section, we'll introduce you to functions that allow you to work with the individual letters within a string.
+You'll learn how to finding the length of a string, extract substrings, and handl long strings in plots and tables.
```
mine-cetinkaya-rundel, line 455:
```suggestion
You'll learn how to find the length of a string, extract substrings, and handle long strings in plots and tables.
```

strings.qmd
```diff
@@ -534,58 +530,135 @@ str_view(str_wrap(x, 30))
...
+The mapping from hexadecimal number to character is called the encoding, and in this case the encoding is called ASCII.
+ASCII does a great job of representing English characters, because it's the **American** Standard Code for Information Interchange.
 
-    This also effects `str_equal()` which can optionally ignore:
+Things aren't so easy for languages other than English.
+In the early days of computing there were many competing standards for encoding non-English characters.
+For example, there were two different encodings for Europe: Latin1 (aka ISO-8859-1) was used for Western European languages and Latin2 (aka ISO-8859-2) was used for Central European languages.
+In Latin1, the byte `b1` is "±", but in Latin2, it's "ą"!
+Fortunately, today there is one standard that is supported almost everywhere: UTF-8.
+UTF-8 can encode just about every character used by humans today, as well as many extra symbols like emoji.
```
mine-cetinkaya-rundel, line 559:
```suggestion
UTF-8 can encode just about every character used by humans today, as well as many extra symbols like emojis.
```

strings.qmd
```diff
@@ -534,58 +530,135 @@ str_view(str_wrap(x, 30))
...
 
-    ```{r}
-    str_equal("i", "I", ignore_case = TRUE)
-    str_equal("i", "I", ignore_case = TRUE, locale = "tr")
-    ```
+readr uses UTF-8 everywhere.
+This is a good default, but will fail for data produced by older systems that don't know use UTF-8.
```
mine-cetinkaya-rundel, line 562:
```suggestion
This is a good default but will fail for data produced by older systems that don't use UTF-8.
```

strings.qmd
```diff
@@ -534,58 +530,135 @@ str_view(str_wrap(x, 30))
...
+If this happens to you, your strings will look weird when you print them.
+Sometimes just one or two characters might be messed up; other times you'll get complete gibberish.
+For example here are two inline csvs with unusual encodings[^strings-8]:
```
mine-cetinkaya-rundel, line 565:
```suggestion
For example here are two inline CSVs with unusual encodings[^strings-8]:
```

strings.qmd
```diff
@@ -534,58 +530,135 @@ str_view(str_wrap(x, 30))
...
+If you're working with individual letters (e.g. with `str_length()` and `str_sub()`) there's an important challenge if you're working with an language that has accents because letters might be represented as an individual character or by combing an unaccented letter (e.g. ü) with a diacritic mark (e.g. ¨).
+For example, this code shows two ways of representing ü that look identical:
 
-[^strings-8]: Sorting in languages that don't have an alphabet (like Chinese) is more complicated still.
+```{r}
+u <- c("\u00fc", "u\u0308")
+str_view(u)
+```
+
+But they have a different length and the first character is different:
```
mine-cetinkaya-rundel, line 611:
```suggestion
But they have different lengths and the first characters are different:
```

strings.qmd
```diff
@@ -534,58 +530,135 @@ str_view(str_wrap(x, 30))
...
+
+```{r}
+str_length(u)
+str_sub(u, 1, 1)
+```
+
+Finally note that these strings look differently when you compare them with `==`, which is stringr provides the handy `str_equal()` function:
```
mine-cetinkaya-rundel, line 618:
```suggestion
Finally note that these strings look differently when you compare them with `==`, for which is stringr provides the handy `str_equal()` function:
```

strings.qmd
```diff
@@ -534,58 +530,135 @@ str_view(str_wrap(x, 30))
...
+str_sort(c("a", "c", "ch", "h", "z"))
+str_sort(c("a", "c", "ch", "h", "z"), locale = "cs")
+```
+
+This also comes up when sorting strings with `dplyr::arrange()` which is why it also has a `locale` argument.
 
 ## Summary
 
-In this chapter you've learned a wide of tools for working with strings, but you haven't learned one of the most important and powerful tools: regular expressions.
+In this chapter you've learned about some of the power of the stringr package: you learned how to create, combine and extract strings, and about some of the challenges you might face with non-English strings.
```
mine-cetinkaya-rundel, line 661:
```suggestion
In this chapter you've learned about some of the power of the stringr package: you learned how to create, combine, and extract strings, and about some of the challenges you might face with non-English strings.
```

mine-cetinkaya-rundel (approved):
All the important bits of the parsing chapter seem to be well covered elsewhere now. My review comments are mostly typo fixes with a couple comments here and there.

One overall comment -- I wonder if we should add to the intro something like: "If you're just starting off it might not be trivial to figure out which chapter a specific function you stumbled upon on StackOverflow or in someone else's code is covered. Use the handy search bar on the top left corner of the online version of the book to find the function in the book."

data-import.qmd
hadley: Yeah when it exists 😄

PR #1130, opened 2022-11-09
Feedback on iteration chapter

jennybc:
The requested review of iteration chapter.

### DIFF
--- a/iteration.qmd
+++ b/iteration.qmd
@@ -476,7 +476,7 @@ Now that we have these 12 paths, we could call `read_excel()` 12 times to get 12
 gapminder_1952 <- readxl::read_excel("data/gapminder/1952.xlsx")
 gapminder_1957 <- readxl::read_excel("data/gapminder/1957.xlsx")
 gapminder_1962 <- readxl::read_excel("data/gapminder/1962.xlsx")
- ...
+ ...,
 gapminder_2007 <- readxl::read_excel("data/gapminder/2007.xlsx")
 ```
 
@@ -619,6 +619,7 @@ In more complicated cases, there might be other variables stored in the director
 In that case, use `set_names()` (without any arguments) to record the full path, and then use `tidyr::separate_wider_delim()` and friends to turn them into useful columns.
 
 ```{r}
+# NOTE: this chapter also depends on dev tidyr (in addition to dev purrr and dev dplyr)
 paths |> 
   set_names() |> 
   map(readxl::read_excel) |> 
@@ -793,7 +794,7 @@ We'll explore this challenge using three examples:
 Sometimes when working with many files at once, it's not possible to fit all your data into memory at once, and you can't do `map(files, read_csv)`.
 One approach to deal with this problem is to load your into a database so you can access just the bits you need with dbplyr.
 
-If you're lucky, the database package will provide a handy function that will take a vector of paths and load them all into the database.
+If you're lucky, the database package you're using will provide a handy function that takes a vector of paths and loads them all into the database.
 This is the case with duckdb's `duckdb_read_csv()`:
 
 ```{r}
@@ -802,9 +803,9 @@ con <- DBI::dbConnect(duckdb::duckdb())
 duckdb::duckdb_read_csv(con, "gapminder", paths)
 ```
 
-This would work great here, but we don't have csv files, we have excel spreadsheets.
+This would work well here, but we don't have csv files, instead we have excel spreadsheets.
 So we're going to have to do it "by hand".
-And learning to do it by hand, will also help you when you have a bunch of csvs and the database that you're working with doesn't have one function that will load them all in.
+Learning to do it by hand will also help you when you have a bunch of csvs and the database that you're working with doesn't have one function that will load them all in.
 
 We need to start by creating a table that will fill in with data.
 The easiest way to do this is by creating a template, a dummy data frame that contains all the columns we want, but only a sampling of the data.
@@ -823,14 +824,14 @@ con <- DBI::dbConnect(duckdb::duckdb())
 DBI::dbCreateTable(con, "gapminder", template)
 ```
 
-`dbCreateTable()` doesn't use the data in `template`, just variable names and types.
+`dbCreateTable()` doesn't use the data in `template`, just the variable names and types.
 So if we inspect the `gapminder` table now you'll see that it's empty but it has the variables we need with the types we expect:
 
 ```{r}
 con |> tbl("gapminder")
 ```
 
-Next, we need a function that takes a single file path and reads it into R, and adds it to the `gapminder` table.
+Next, we need a function that takes a single file path, reads it into R, and adds the result to the `gapminder` table.
 We can do that by combining `read_excel()` with `DBI::dbAppendTable()`:
 
 ```{r}
@@ -842,7 +843,7 @@ append_file <- function(path) {
 }
 ```
 
-Now we need to call `append_csv()` once for `path`.
+Now we need to call `append_csv()` once for each element of `paths`.
 That's certainly possible with `map()`:
 
 ```{r}
@@ -857,7 +858,7 @@ But we don't care about the output of `append_file()`, so instead of `map()` it'
 paths |> walk(append_file)
 ```
 
-Now if we can see we have all the data in our table:
+Now we can see if we have all the data in our table:
 
 ```{r}
 con |> 
@@ -866,15 +867,16 @@ con |>
 ```
 
 ```{r, include = FALSE}
+# why not show this?
 DBI::dbDisconnect(con, shutdown = TRUE)
 ```
 
 ### Writing csv files
 
 The same basic principle applies if we want to write multiple csv files, one for each group.
-Let's imagine that we want to take the `ggplot2::diamonds` data and save our one csv file for each `clarity`.
+Let's imagine that we want to take the `ggplot2::diamonds` data and save one csv file for each `clarity`.
 First we need to make those individual datasets.
-There are many ways you could that, but there's one way we particularly like: `group_nest()`.
+There are many ways you could do that, but there's one way we particularly like: `group_nest()`.
 
 ```{r}
 by_clarity <- diamonds |> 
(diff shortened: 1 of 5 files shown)


iteration.qmd
```diff
@@ -29,7 +29,8 @@ If you want to live life on the edge you can get the dev version with `devtools:
 :::
 
 In this chapter, we'll focus on tools provided by dplyr and purrr, both core members of the tidyverse.
-You've seen dplyr before, but purrr is new.
+You've seen dplyr before, but purrr LINK?
```
jennybc, line 32: Is there some existing convention in the book that would have you linking to <https://purrr.tidyverse.org> or similar, upon first mention?

iteration.qmd
```diff
@@ -76,13 +77,13 @@ df |> summarise(
 
 `across()` has three particularly important arguments, which we'll discuss in detail in the following sections.
 You'll use the first two every time you use `across()`: the first argument, `.cols`, specifies which columns you want to iterate over, and the second argument, `.fns`, specifies what to do with each column.
-You also the `.names` argument when you need additional control over the output names, which is particularly important when you use `across()` with `mutate()`.
+You can use the `.names` argument when you need additional control over the names of output columns, which is particularly important when you use `across()` with `mutate()`.
```
jennybc, line 80: In general it feels like there's a lot of switching between "columns" and "variables" FWIW.

iteration.qmd
```diff
@@ -168,7 +169,10 @@ df |>
 
 This is a little verbose, so R comes with a handy shortcut: for this sort of throw away, or **anonymous**[^iteration-1], function you can replace `function` with `\`:
 
-[^iteration-1]: Anonymous, because didn't give it a name with `<-.`
+[^iteration-1]: Anonymous, because we never explicitly created it and gave it a name with `<-`.
+    Another term programmers use for this is "lambda function".
```
jennybc, line 173: I like giving people this sort of background, i.e. de-mystifying jargon CS people love to throw around.

iteration.qmd
```diff
@@ -168,7 +169,10 @@ df |>
...
+
+    MAYBE: explain the older syntax `~` in case people see it?
```
jennybc, line 175: I have no idea if you're intentionally not mentioning `~`.

iteration.qmd
```diff
@@ -246,18 +250,19 @@ df |>
   )
 ```
 
-If you'd like to instead create new columns, you can use the `.names` argument give the output new names:
+If you'd like to instead create new columns, you can use the `.names` argument to give the output new names:
 
 ```{r}
 df |> 
   mutate(
-    across(a:d, \(x) x * 2, .names = "{.col}_double")
```
jennybc, line 254: I assume they've seen "double" in the context of double precision floating point numbers, so seems best to not risk confusion. Just do something else.

iteration.qmd
```diff
@@ -285,9 +290,18 @@ expand_dates <- function(df) {
...
 }
+
+df2 <- tibble(
+  name = c("Amy", "Bob"),
+  date = as.Date(c("2009-08-03", "2010-01-16"))
+)
+
+df2 |> 
+  expand_dates()
+# I think deserves to be applied to some actual data, i.e. complete the example. Most readers won't be able to imagine what the `list()` thing does. And I guess my immediate attempt to use it has failed?
```
jennybc, line 301: I think this example should be taken to its logical conclusion, i.e. `expand_dates()` should be shown "in action".

iteration.qmd
```diff
@@ -285,9 +290,18 @@ expand_dates <- function(df) {
...
 ```
 
-`across()` also makes it easy to supply multiple variables in a single argument because the first argument uses tidy-select; you just need to remember to embrace that argument.
+`across()` also makes it easy to supply multiple variables in a single argument because the first argument uses tidy-select; you just need to remember to embrace that argument LINK.
```
jennybc, line 304: I assume there's some where to link to re: embracing.

iteration.qmd
```diff
@@ -310,7 +324,7 @@ diamonds |>
 
 ### Vs `pivot_longer()`
 
-Before we go on, it's worth pointing out an interesting connection between `across()` and `pivot_longer()`.
+Before we go on, it's worth pointing out an interesting connection between `across()` and `pivot_longer()` LINK.
```
jennybc, line 327: Is there a place to link to re: `pivot_longer()`?

jennybc:
General observation (I hit it twice when working the exercises for the "multiple columns" content): If someone really leans into the IDE's auto-completion, you have to fight it a bit to get naked functions, as you might need for `where()` and `across(.fns =)`.

example with `.fns`

``` r
palmerpenguins::penguins |> 
  summarise(across(everything(), n_distinct()))
#> Error in `summarise()`:
#> ! Problem while computing `..1 = across(everything(), n_distinct())`.
#> Caused by error in `across()`:
#> ! `.fns` must be NULL, a function, a formula, or a list of
#>   functions/formulas.

#> Backtrace:
#>     ▆
#>  1. ├─dplyr::summarise(...)
#>  2. └─dplyr:::summarise.data.frame(...)
#>  3.   └─dplyr:::summarise_cols(.data, dplyr_quosures(...), caller_env = caller_env())
#>  4.     ├─base::withCallingHandlers(...)
#>  5.     └─dplyr:::expand_across(dots[[i]])
#>  6.       └─dplyr:::across_setup(...)
#>  7.         └─rlang::abort(msg, call = call(across_if_fn))

palmerpenguins::penguins |> 
  summarise(across(everything(), n_distinct))
#> # A tibble: 1 × 8
#>   species island bill_length_mm bill_depth_mm flipper_length_mm body_…¹   sex  year
#>     <int>  <int>          <int>         <int>             <int>   <int> <int> <int>
#> 1       3      3            165            81                56      95     3     3
#> # … with abbreviated variable name ¹​body_mass_g
```

example inside `where()`

``` r
diamonds |> 
  group_by(cut, clarity, color) |> 
  summarise(across(where(is.numeric()), mean))
```

It's interesting to think about whether there's anything the IDE could do? Or is it feasible to throw a better error? Both sound pretty hard.

In which case, maybe it's worth showing this gotcha once in this chapter, so readers are better at recognizing what went wrong.

iteration.qmd
```diff
@@ -421,18 +441,19 @@ data <- bind_rows(data2019, data2020, data2021, data2022)
...
 You can imagine that this would get tedious quickly, especially if you had hundreds of files, not just four.
 The following sections show you how to automate this sort of task.
-There are three basic steps: use `dir()` list all the files in a directory, then use `purrr::map()` to read each of them into a list, then use `purrr::list_rbind()` to combine them into a single data frame.
+There are three basic steps: use `dir()` to list all the files in a directory, then use `purrr::map()` to read each of them into a list, then use `purrr::list_rbind()` to combine them into a single data frame.
 We'll then discuss how you can handle situations of increasing heterogeneity, where you can't do exactly the same thing to every file.
 
 ### Listing files in a directory
 
 `dir()` lists the files in a directory.
+TO CONSIDER: why not use it via the more obvious name `list.files()`?
```
jennybc, line 450: If we're not going to plug fs, then I think I'd use `list.files()` over `dir()`.

iteration.qmd
jennybc: It's done for targets, for example.

iteration.qmd
```diff
@@ -322,21 +336,25 @@ df |>
     median = median(value, na.rm = TRUE),
     n_miss = sum(is.na(value))
   )
+
+# this feels unfinished, i.e. lacking the direct head-to-head comparison to something using across()
```
jennybc, line 340: In particular, I think there's another pivot operation that will happen and it would be get to really nail down that pattern, so the reader knows that's normal.

iteration.qmd
```diff
@@ -772,7 +794,7 @@ We'll explore this challenge using three examples:
 Sometimes when working with many files at once, it's not possible to fit all your data into memory at once, and you can't do `map(files, read_csv)`.
 One approach to deal with this problem is to load your into a database so you can access just the bits you need with dbplyr.
 
-If you're lucky, the database package will provide a handy function that will take a vector of paths and load them all into the database.
+If you're lucky, the database package you're using will provide a handy function that takes a vector of paths and loads them all into the database.
```
jennybc, line 797: Without this, it felt like maybe there was a package named "database".

iteration.qmd
```diff
@@ -845,15 +867,16 @@ con |>
 ```
 
 ```{r, include = FALSE}
+# why not show this?
```
jennybc, line 870: I lean towards showing clean up code.

iteration.qmd
```diff
@@ -951,7 +976,7 @@ ggsave(by_clarity$path[[8]], by_clarity$plot[[8]], width = 6, height = 6)
 
 ```{r}
 #| include: false
-unlink(by_clarity$paths)
+unlink(by_clarity$path)
```
jennybc, line 979: Same comment about making this visible.

jennybc (commented):
I've made lots of inline suggestions or comments.

I've got two more "meta" comments.

---

Perhaps there could be even more about how iteration works in R vs. other languages. There's a bit of this already in the intro, but I think you could lean into that even more, in terms of helping these readers make sense of (and debunk) some typical criticisms of R. I think it helps people understand why reading and writing R code feels different from re-purposing an all-purpose programming language for data science.

As you say, we get lots of iteration "for free", simply because of R's default posture re: vectorization. What's less obvious but also true is that R's whole orientation towards data analysis changes how we iterate (at least, *it can*). R + tidyverse offers some great idioms for doing something to some or all columns and for doing something to every group. Therefore, where folks writing in other languages and coming from other programming traditions would use a for loop? We create or rectangle our way to a data structure where the iteration problem can be nicely solved with our existing column- or group-wise toolkit.

The "heterogeneous data" section is a great example of turning the data exploration/cleaning into a data science task.

---

I wonder if you should also add some explicit statement about "functional programming". Again, in the name of demystifying jargon. I assume this is the first or main place in the book where these readers are exposed to the practice of passing a function to a function.

iteration.qmd
hadley: I've standardised on columns

iteration.qmd
hadley: Because it doesn't actually quiet the message I want it to; so there's something else going on here that I don't understand.

hadley:
Thanks @jennybc! This was super useful

Issue #1050, opened 2022-06-24
Consider chapter on binary data formats


hadley:
i.e. arrow + parquet

Useful content at https://arrow-user2022.netlify.app

* [ ] explore nyctaxi data in `_backup-exclude`
* [ ] convert to parquet
* [ ] show how to use with arrow
* [ ] show how to use with duckdb
* [ ] a couple of benchmarks showing how fast it is

hadley:
Maybe something like the following as a rough overview. Would appear after the databases chapter, I think.

Introduction to arrow
* out of memory, fast
* cross-language

The nyctaxi dataset
* how to get the csv files (the pain of big data)
* show how to convert to parquet
* something general about the multiple files strategy
* mention arrow (e.g.) S3 support

arrow + dplyr
* it's just dplyr
* collect()
* things that don't work (yet)

duckdb
* show how you can also access via duckdb and dbplyr
* keep it brief; point is to illustrate using same data source in multiple tools

PR #1115, opened 2022-10-30
New visualize part

mine-cetinkaya-rundel:
Visualize related updates:

- [x] Rewrite the whole game's data visualization chapter to include a higher-level overview using palmerpenguins.
- [x] Create a 3 chapter visualize part: layers, eda, communication.
- [x] Write layers, mostly based on old whole game / data visualization chapter.
- [x] Move EDA out of whole game and into new visualize part, removing bits of it used in whole game's data visualization chapter.
- [x] Move graphics for communication out of communicate part and into here.
- [x] Add example from scales package.
- [x] Add layout/patchwork to new comminucation.
- [ ] Find a better name for communication chapter to avoid confusion with communicate part on Quarto.
- [x] Add alt text to each figure.

PR #1137, opened 2022-11-16
adds binary data chapter

djnavarro:
This is a first pass at the binary data chapter proposed in #1050. The structure isn't precisely identical to the proposed version, and currently looks like this:

- the arrow package
    - uses the Seattle library data to illustrate `open_dataset()` 
    - illustrates dplyr workflows with Dataset objects
- the parquet file format
    - illustrates limitations of the csv plain text format
    - describes the parquet format
    - shows use of `read_parquet()` and `write_parquet()`
    - demonstrates parquet format advantages
- multi-file data sets
    - illustrates use of `write_dataset()`
    - gives guidelines about partitioning data
    - describes hive partitioning conventions
    - demonstrates performance advantages of well-partitioned data
 
There is a section at the end with content for which I've sketched an initial draft but am not sure where to integrate into the text:

- using duckdb with arrow: shows use of `to_duckdb()` 
- the arrow in-memory and on-disk formats: briefly discusses Table objects and the arrow file format as the on-disk analog
- data in the cloud: briefly highlights `s3_bucket()` and `gs_bucket()` using the S3 copy of the diamonds data set

I hope it goes without saying that this is a rough first draft, so please feel free to offer any and all suggestions. Tagging folks from the email thread: @hadley, @nealrichardson, @jonkeane, @thisisnic, @stephhazlitt 🙂


binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+status("drafting")
+```
+
+## Introduction
+
+### Prerequisites
+
+This chapter focuses on large data sets, and binary file formats that are well suited to working with them.
+We'll use a data set of item checkouts from Seattle public libraries, available online at [data.seattle.gov/Community/Checkouts-by-Title/tmmm-ytt6](https://data.seattle.gov/Community/Checkouts-by-Title/tmmm-ytt6).
+The data is distributed as a single 9GB csv file, so it may take some time to download.
```
hadley, line 16: Should we make this available in an S3 bucket (or similar)? And maybe suggest some R code to download it?

Maybe the first section in the chapter should be "getting big datasets", where you could show how to download a file with code and then maybe move in the contents of "Data in the cloud"

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
+# Binary data formats {#sec-binary-data}
```
hadley, line 1: Maybe we should freeze (https://quarto.org/docs/projects/code-execution.html#freeze) this doc so we don't need to try and rebuild on github?

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+)
+```
+
+When this code is run, `open_dataset()` scans the file to work out its structure and creates the `seattle_lib` object that stores metadata.
+The data remain on-disk, and are not loaded into memory.
+Nevertheless, we are able to call dplyr functions to interact with the data:
+
+```{r glimpse-data}
+#| cache: true
+glimpse(seattle_lib)
```
hadley, line 56: What does it look like if you just print it?

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+## The arrow package
+
+In this chapter we'll rely on Apache Arrow ([arrow.apache.org](https://arrow.apache.org/)), a multi-language toolbox designed for efficient analysis and transport of large data sets.
+The arrow package provides a standard way to use Apache Arrow in R, and allows you to analyze larger-than-memory datasets using familiar dplyr syntax.
+As an additional benefit, arrow is extremely fast: you'll see some examples later in the chapter.
+
+### Opening a dataset
+
+Let's start by taking a look at the Seattle library checkout data.
+At 9GB in size, the csv file is large enough that we probably don't want to load the whole thing into memory.
```
hadley, line 37: Mention some rule of thumb like that you usually want 2x the memory?

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+#| cache: true
+seattle_lib |> 
+  filter(MaterialType == "BOOK") |>
+  group_by(CheckoutYear) |>
+  summarise(TotalCheckouts = sum(Checkouts)) |>
+  arrange(desc(CheckoutYear)) |>
+  collect()
+```
+
+Only some R expressions are understood by arrow, so you will not necessarily be able to write exactly the same code you usually would.
```
hadley, line 78: Like dbplyr, ...

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+However, the list of operations and functions supported is fairly extensive and growing.
+A complete list is provided in the arrow package documentation at [arrow.apache.org/docs/r/reference/acero.html](https://arrow.apache.org/docs/r/reference/acero.html).
+
+## The parquet file format
+
+When working with small data sets, csv files can be convenient: they're human readable plain text files, and most analysis tools can read data from a csv.
+However, there are some drawbacks:
+
+-   Ambiguity.
+    A csv file does not uniquely specify the data structure.
```
hadley, line 88: Need to link to "column types" section (currently only available in https://github.com/hadley/r4ds/pull/1128)

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+
+    ```{r csv-ambiguity-failure}
+    csv_file <- tempfile(fileext = ".csv")
+    write_csv(original, csv_file)
+    read_csv(csv_file, show_col_types = FALSE)
+    ```
+
+    The `read_csv()` and `write_csv()` functions are designed to avoid this issue to an extent, but it is impossible to avoid it completely because the csv format is inherently ambiguous.
+
+-   Inaccuracy.
```
hadley, line 114: I'd make this the last bullet, since it'd be unusual for this to make any practical difference.

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+      count(same_value)
+    ```
+
+    A substantial proportion of the reloaded values differ from the original ones.
+    These differences are very small in magnitude, and occur because the csv represents numeric values with limited precision.
+    Even so this can still be a source of problems in your analysis code.
+
+-   Poor performance.
+    Compared to other formats, csv files can be slow to read and write because the on-disk format is very different to how the data needs to be represented in-memory.
+    We'll see some examples of this later in the chapter.
```
hadley, line 141: Need "Large size" bullet to talk about (lack of) compression

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+#| cache: true
+glimpse(seattle_lib)
+```
+
+The first line in the output tell you that `seattle_lib` is stored locally on-disk as a single csv file.
+The second line indicates that the data are stored as a tabular data set with 41 million rows and 12 columns.
+
+### Using dplyr with arrow
+
+Writing dplyr code for arrow data is conceptually similar to dbplyr, described in @sec-import-databases.
```
hadley, line 64: I think this section could do with a bit more content before diving into the technical details. I think that maybe you could save to partitioned parquet here to talk a bit about performance, and then go into the details later.  (i.e move some of the "Parquet files are faster" section here).

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+    collect()
+})
+```
+
+The analysis of the parquet data finishes in about one tenth the time it takes to perform the same analysis with the csv data.
+
+## Multi-file data sets
+
+When data sets become large it is often useful to split the data across many files rather than storing everything in a single large file.
+When this structuring is done intelligently, the multi-file strategy can lead to significant improvements in performance.
```
hadley, line 271: Because it allows multiple cores to work simultaneously?

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+
+This is a 10-fold speed up compared to a single parquet file, and a 100-fold speed up compared to the original csv file.
+
+## Other topics
+
+I haven't quite worked out the best way to integrate these sections.
+
+### Using duckdb with arrow
+
+One useful feature of arrow is that it is interoperable with duckdb and dbplyr.
```
hadley, line 346: Maybe this could go be a subsection in "Using dplyr with arrow"? With some connection like:

> There's also an even more direct connection between arrow and dbplyr — it's very easy to turn an arrow dataset into a duckdb datasource. This doesn't involve any memory copying, and speaks to the goals of the arrow ecosystem: enable seamless transitions from one computing framework to another.

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+  group_by(CheckoutMonth) |>
+  summarise(TotalCheckouts = sum(Checkouts)) |>
+  arrange(desc(CheckoutMonth)) |>
+  collect()
+```
+
+There is an analogous function `to_arrow()` that passes control back to arrow, meaning that it is possible to swap back and forth between the duckdb and arrow back ends within a single analysis pipeline.
+This operation is extremely efficient (only the metadata need to be copied from one system to the other) can occasionally be valuable when one system implements functionality that the other does not.
+
+### The arrow in-memory and on-disk format
```
hadley, line 363: I think we could probably just skip this section?

hadley (commented):
Thanks for working on this!!!

thisisnic (commented):
Is it also worth mentioning something about specifying schemas when opening CSV datasets, or including an example of this?  This is something we've ended up with a lot of Stack Overflow + JIRA questions about in the past.

binary-data-formats.qmd
nealrichardson: The print method prints column names and types (i.e. the schema) but does not print any data values.

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+      mutate(same_value = old_x == new_x) |>
+      count(same_value)
+    ```
+
+    A substantial proportion of the reloaded values differ from the original ones.
+    These differences are very small in magnitude, and occur because the csv represents numeric values with limited precision.
+    Even so this can still be a source of problems in your analysis code.
+
+-   Poor performance.
+    Compared to other formats, csv files can be slow to read and write because the on-disk format is very different to how the data needs to be represented in-memory.
```
nealrichardson, line 140: Modern CSV readers get decent read speed, though not as good as with binary formats. But the bigger difference is in file size, which does contribute to poor speed depending on the speed of your storage.

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+    We'll see some examples of this later in the chapter.
+
+The Apache Parquet file format ([parquet.apache.org](https://parquet.apache.org/)) is designed to address these limitations.
+Like a csv file, a parquet file stores a single rectangular data set.
+Features of the parquet format that make it appealing for data scientists include:
+
+-   Parquet files store data in a binary format (not plain text), avoiding inaccuracies in encoding and making it faster to read and write than a csv file.
+-   Parquet files store metadata that uniquely identifies the data types stored in each column, which avoids the ambiguity of csv files.
+-   Parquet files are compressed by default, and are typically much smaller than the corresponding csv files would be.
```
nealrichardson, line 149: "default" is a library setting and I don't know that all parquet writers compress by default. But compression is supported and commonly used. 

The other way parquet gets better file sizes is with encodings: https://parquet.apache.org/docs/file-format/data-pages/encodings/

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+-   Parquet files internally break the data into distinct "chunks" each with associated metadata, making it possible to write intelligent file readers that can jump straight to a specific part of the data when the full data set is not needed.
```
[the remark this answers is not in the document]
nealrichardson: While I do see the word "chunk" on https://parquet.apache.org/docs/file-format/, I would think that if there is a term about this that would need quoting, it would be "row groups". Maybe it's best just to leave `chunk` as a word and not quoted like it is an official term.

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+
+All these factors together make the parquet format appealing when working with larger data sets.
```
nealrichardson, line 152: Another point worth mentioning: Parquet is widely used by big data systems, so adopting it doesn't mean locking yourself or your collaborators into an R only workflow.

binary-data-formats.qmd
nealrichardson: `arrow` does multithreaded reading with single files too

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+There are no hard and fast rules about how to partition a data set: the results will depend on your data, access patterns, and the systems that read the data.
+As a rough guide, the arrow package documentation suggests the following heuristics:
+
+-   Avoid files smaller than 20MB and larger than 2GB.
+-   Avoid partitioning layouts that produce more than 10,000 distinct files.
+
+According to these guidelines, the Seattle library data is a little too large to be stored as a single 4GB parquet file.
+We should consider splitting it up.
+
+As a general rule, a useful strategy for partitioning a large dataset is to create meaningful subsets based on a variable that you're likely to query on.
```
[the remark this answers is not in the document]
nealrichardson: Not just "query" but especially "filter". You *might* get some benefits when aggregating on partition columns, but the big benefit is predicate pushdown with your filters so you don't have to touch entire files.

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+  to_duckdb() |>
+  filter(CheckoutYear == 2021, MaterialType == "BOOK") |>
+  group_by(CheckoutMonth) |>
+  summarise(TotalCheckouts = sum(Checkouts)) |>
+  arrange(desc(CheckoutMonth)) |>
+  collect()
+```
+
+There is an analogous function `to_arrow()` that passes control back to arrow, meaning that it is possible to swap back and forth between the duckdb and arrow back ends within a single analysis pipeline.
+This operation is extremely efficient (only the metadata need to be copied from one system to the other) can occasionally be valuable when one system implements functionality that the other does not.
```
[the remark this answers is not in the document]
nealrichardson:
```suggestion
This operation is extremely efficient because both libraries deal with the Arrow format, so data is passed between then without making a copy. This can occasionally be valuable when one tool implements functionality that the other does not.
```

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+
+```{r parquet-precision-read}
+reloaded <- read_parquet(parquet_file) |>
+  rename(new_x = old_x)
+```
+
+Finally we bind the old and new data sets together and compare them:
+
+```{r parquet-precision-success}
+bind_cols(original, reloaded) |>
```
nealrichardson, line 202: Is it worth noting in the read/write_parquet examples here that you're dealing with R data.frames? `bind_cols()` isn't supported on arrow Tables or Datasets, it only works here because `read_parquet()` defaults to reading in a data.frame.

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+
+Finally we bind the old and new data sets together and compare them:
+
+```{r parquet-precision-success}
+bind_cols(original, reloaded) |> 
+  mutate(same_value = old_x == new_x) |>
+  count(same_value)
+```
+
+As you can see, numeric values stored in parquet files are stored faithfully.
```
nealrichardson, line 207: Perhaps worth noting (maybe above when you talk about the benefits of Parquet) that Parquet (and Arrow for that matter) support more types than R does. So you can get better precision and support operations (like calculations on decimal types) that are not possible in R. In fact, you may get a loss of fidelity reading some types into R.

binary-data-formats.qmd
hadley: My qualm with glimpse is that I don't think we use it elsewhere in the book, but maybe I should fix that.

mine-cetinkaya-rundel: I use `glimpse()` pretty regularly in teaching, so I think it's good to add to other places in the book.

binary-data-formats.qmd
jonkeane: Agreed with this — adding details of where | how to get this would be fantastic. And if we presume that someone will already have (or be willing to install) arrow in this chapter, we could utilize [`copy_files()`](https://arrow.apache.org/docs/r/reference/copy_files.html) which IMHO is a pretty nice way to fetch files from S3

binary-data-formats.qmd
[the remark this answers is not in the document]
jonkeane: I very much agree with dropping the quotes on `chunk` here too. I'm probably more allergic than most to quotes in these circumstances, but even if someone doesn't totally grok the technical details of row groups and how those are chunks, the basic concept of chunks and chunking is applicable enough here to facilitate understanding (i.e. we're not using the word in a super specific or super novel way that we need to flag it as special like this IMHO)

binary-data-formats.qmd
hadley: @jonkeane does that have a progress bar? and some ability to resume if the connection is interrupted?

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+library(arrow)
+library(duckdb)
+library(dbplyr)
+```
+
+## The arrow package
+
+In this chapter we'll rely on Apache Arrow ([arrow.apache.org](https://arrow.apache.org/)), a multi-language toolbox designed for efficient analysis and transport of large data sets.
+The arrow package provides a standard way to use Apache Arrow in R, and allows you to analyze larger-than-memory datasets using familiar dplyr syntax.
+As an additional benefit, arrow is extremely fast: you'll see some examples later in the chapter.
```
jonkeane, line 32: Would this be a good place to put a direct link to the arrow R package docs too? I know it's "just" two clicks away at arrow.apache.org, but for some folks it might be just disorienting enough to be impactful?

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+Instead, we'll use `arrow::open_dataset()` to scan the file on-disk and create an Arrow Dataset object called `seattle_lib`:
+
+```{r open-dataset}
+seattle_lib <- open_dataset(
+  sources = "data/seattle-library-checkouts.csv", 
+  format = "csv"
+)
+```
+
+When this code is run, `open_dataset()` scans the file to work out its structure and creates the `seattle_lib` object that stores metadata.
```
jonkeane, line 50: You might have decided to tactically leave out this detail, but I wonder if we should still mention that "scans" in this sentence is only the first little bit of the file (enough to do type inference and construct a schema) and is not the whole file?

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+
+## The parquet file format
+
+When working with small data sets, csv files can be convenient: they're human readable plain text files, and most analysis tools can read data from a csv.
+However, there are some drawbacks:
+
+-   Ambiguity.
+    A csv file does not uniquely specify the data structure.
+    For example, a csv reader has to guess whether `"2022-10-08"` should be parsed as a string or a date.
+    If it is a date, it has to guess whether the date refers to October 8th or August 10th.
```
jonkeane, line 90: `YYYY-DD-MM` is prue date format chaos. I'm fine to leave this as is, but maybe using `08-10-2022` would elucidate the point with an example that many people have had to deal with?

binary-data-formats.qmd
jonkeane: Yeah, I'll piggyback on this comment to also say that it might be worthwhile to explicitly point out that though Apache Parquet is binary, it is an open standards-based format (as well as it being fairly widely adopted like Neal points out)

I will admit when I first encountered parquet I was personally hesitant because I thought "ugh some new proprietary binary format that I'll be stuck locked into". Even with the apache moniker in the intro, I'm not sure that would be a shibboleth of openness to everyone who is reading R4DS like it is to us.

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+  ) |>
+    filter(CheckoutYear == 2021, MaterialType == "BOOK") |>
+    group_by(CheckoutMonth) |>
+    summarise(TotalCheckouts = sum(Checkouts)) |>
+    arrange(desc(CheckoutMonth)) |>
+    collect()
+})
+```
+
+This is a 10-fold speed up compared to a single parquet file, and a 100-fold speed up compared to the original csv file.
```
jonkeane, line 338: I love this quantification and it's pretty nice that they are roundish numbers. Jokingly: it's beginning to sound like a tech product launch "And with our new MPUs, we've got 100x flops" :joy

binary-data-formats.qmd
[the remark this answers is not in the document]
jonkeane: If you take Neal's suggestion of 

```
This operation is extremely efficient because both libraries deal with the Arrow format, so data is passed between then without making a copy. This can occasionally be valuable when one tool implements functionality that the other does not.
```

We might also add in something about serialization as well as the copying bits. Both are important, but in some ways not needing to (re)encode the data to send it is the biggest time saver

binary-data-formats.qmd
```diff
@@ -0,0 +1,404 @@
...
+arrow_file <- tempfile(fileext = ".arrow")
+write_ipc_file(checkouts2021, arrow_file)
+```
+
+Note that for future releases of arrow, these functions are likely to deprecated in favour of the names `read_arrow()` and `write_arrow()` respectively.
+Also note that the arrow file format was previously known as the "feather" file format, and you are likely to see references to it by that name.
+
+To learn more about Datasets, Tables, and other arrow data objects, see the documentation on the arrow package website: [arrow.apache.org/docs/r/](https://arrow.apache.org/docs/r/).
+
+### Data in the cloud
```
jonkeane, line 392: If we keep this section (personally, I would vote yes, but not sure what the tradeoffs are at this point...) it would be nice to also talk about how the connection of datasets being able to query only the parts of parquet files that it needs to + S3 makes for a shockingly efficient setup for processing data out of S3.

binary-data-formats.qmd
jonkeane: IMO, it should do both — though AFAIK it doesn't currently.

We have an issue for a progress bar: https://issues.apache.org/jira/browse/ARROW-14611 (though there was some hesitancy there that it might be out of scope for the Arrow C++ internals)

So maybe ultimately we should use a different method of fetching these that has better UX than `copy_files` currently does?

hadley: Yeah, that would be nice, but I don't know of anything off the rack 😞

binary-data-formats.qmd
hadley: Could you expand on this a bit? Are you saying that arrow will read just the necessary bytes from the parquet header so it can only read the columns you request? Presumably it can only return a subset of rows the rows if the parquet file is sharded by the variable you're filtering on?

But doesn't this all have to happen a second time if you re-run the code? Or does arrow somehow magically cache the data locally?

binary-data-formats.qmd
djnavarro: +1 for using glimpse a little more in the book. I use glimpse a lot in writing posts and (used to, when I taught) find it really helpful for teaching purposes too. Happy to remove it here though if that's preferable

binary-data-formats.qmd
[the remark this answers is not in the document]
djnavarro: Quote marks dropped :-)

binary-data-formats.qmd
mine-cetinkaya-rundel: I've added it in #1115, I'd say should be good to leave in

hadley: I also added it to https://r4ds.hadley.nz/data-transform.html#nycflights13

binary-data-formats.qmd
jonkeane: Yeah to most of what you said:

> Are you saying that arrow will read just the necessary bytes from the parquet header so it can only read the columns you request?

Yeah, one of the parts of registering a parquet dataset in S3 is reading the headers of (some of) the files (it will read all of them if you tell it to unify schemas, otherwise it only reads the first — or perhaps the first of each shard). Then when you go to actually scan the files, it does again read the header(s) and grabs only the columns that you specify.

> Presumably it can only return a subset of rows the rows if the parquet file is sharded by the variable you're filtering on?

If the Parquet file has its rows split up into row groups (which Arrow will do by default if the files are large enough), it can use the row group statistics to discard whole chunks of row groups at a time if they don't meet a specific filter. Like you allude to, that means you'll only get a benefit if there are row groups that one can ignore — which might take careful writing of the data in the first place like with sharding. Though I've found it very useful in cases where I'm filtering out rows that contain `NA`s|`NULL`s which happen to be distributed unevenly throughout the dataset (but frequently enough that many whole row groups will be all `NA` for that column and thus can be ignored.

As for cache: I don't know of Arrow doing anything special here — so yeah if you query again it'll do the whole thing over. This workflow is good for the kinds of one-shot queries and then having something local in memory that you work with after that, but not for repeated querying of the dataset like that.

hadley: Thanks for the explanation!

binary-data-formats.qmd
stephhazlitt: +1 for starting the chapter with getting some large (or subset of large) data. If the `copy_files` UX is too minimal (however it is my go-to for S3/GCS at this time), there is the option of letting the reader manually download the data _or_ point `open_dataset()` to the example data hosted in an S3 bucket. There are trade-offs -- the latter can be _slow_ but also means not downloading anything.

stephhazlitt:
So exciting to see this chapter coming together @djnavarro 🙌. I wonder about leveraging some of your work+iterations currently happening on the arrow R package docs [ARROW-17887](ARROW-17887). I really like the Arrow tools for in-memory and Arrow tools for larger-than-memory narrative. All the same pieces are here, just in a different order. What about:

- [ ] what is Arrow
- [ ] get some largish data (or a subset), including from the cloud
- [ ] in-memory (Parquet format benefits, read+write)
- [ ] larger-than-memory (too large single files 😢, multi-file benefits, read+write)
- [ ] manipulating Datasets with dplyr+Arrow C++ engine
- [ ] manipulating Datasets with duckdb
- [ ] where to read more

binary-data-formats.qmd
djnavarro: Maybe? I'm wondering if maybe I find a different way to write the code snippet. My worry here is that the brief for the chapter suggested by @hadley focuses on parquet primarily. I'm trying to bring in only a bare minimum of arrow concepts, so perhaps the best thing to do here is pick examples where Tables, Datasets, and data frames all behave the same way? Hopefully when https://github.com/apache/arrow/pull/14514 merges it will be a little easier for us to link out to the Arrow docs here in a way that doesn't confuse Arrow novices?

binary-data-formats.qmd
djnavarro: yeah I was thinking about partitioning. I've reworded to hint to the reader that the advantage comes from reading only a subset of files. Trying to be slightly vague here because it doesn't exactly feel like the spot to start talking about predicate pushdown!

binary-data-formats.qmd
djnavarro: Given the concerns expressed about the intended scope of the arrow FileSystem interface, and the intended audience of R4DS, it might be best to avoid using `copy_files()` in this chapter; and relatedly, I'm now leaning towards replacing the section talking about Arrow cloud support with a more generic sentence that mentions that this support exists and then links out to the relevant vignette in the Arrow docs.

In the meantime, I'm wondering if the best approach is to host a copy of the csv file in an s3 bucket and then fall back on `download.file()` as the example of how to download?

binary-data-formats.qmd
djnavarro: I guess what I'm wondering with this is how much is worth saying in the R4DS chapter and how much do we want to push into the Arrow docs? To me it feels like most of @jonkeane's explanation is something that should be added into the "working with cloud storage" vignette on the Arrow site. I don't know if we want to be going that deep in R4DS? (Probably @hadley's call!) 

I guess the prpblem at the moment is that on the Arrow docs it's not at all clear to users that this advantage exists, and (unfortunately) this isn't one of the vignettes I worked much on in https://github.com/apache/arrow/pull/14514/. I could open a ticket to expand on the vignette there perhaps?

binary-data-formats.qmd
```diff
@@ -0,0 +1,236 @@
...
+glimpse(seattle_lib)
+```
+
+The first line in the output tell you that `seattle_lib` is stored locally on-disk as a single csv file.
+The second line indicates that the data are stored as a tabular data set with 41 million rows and 12 columns.
+
+### Partitioning datasets
+
+When data sets become large it is often useful to split them across many files rather than storing everything in a single large file.
+When this structuring is done intelligently, the multi-file strategy can lead to significant improvements in performance because some analyses will only require a subset of the files to be read.
```
hadley, line 66: And you your computer can work on multiple pieces in parallel? Or is that true if you use parquet even without partitioning?

binary-data-formats.qmd
```diff
@@ -0,0 +1,236 @@
...
+The `path` argument is used to specify the folder into which all the data files will be written, and the `format` argument is used to specify the file format for the data files.
+We'll talk more about the parquet format in @sec-parquet.
+For the moment, let's take a look at the files:
+
+```{r show-parquet-files}
+seattle_lib_files <- list.files(
+  path = "data/seattle-library-checkouts",
+  recursive = TRUE
+) 
+seattle_lib_files
```
hadley, line 107: I wonder if it's worth turning this into a tibble? Then you could have a column for file size (in MB?) and it would automatically truncate to just the first six files.

hadley (commented):
This looks to be very close to done to me. I have a couple last comments/questions below and then I think we can merge 😀

binary-data-formats.qmd
djnavarro: I am pretty sure this is true for parquet data even without partitioning because the Dataset API can leverage the internal row grouping within the parquet file. I can't verify the difference using a dplyr (because honestly I haven't quite worked out all the details of how the dplyr code translates!), but here's an example using the low level Scanner API:

``` r
library(arrow, warn.conflicts = FALSE)
seattle_lib <- open_dataset("data/seattle-library-checkouts.parquet")

multi_threaded_scan <- Scanner$create(
  dataset = seattle_lib,
  filter = Expression$field_ref("CheckoutYear") == 2021,
  use_threads = TRUE
)
system.time(multi_threaded_scan$ToTable())
#>    user  system elapsed 
#>  22.341   6.826   7.924

single_threaded_scan <- Scanner$create(
  dataset = seattle_lib,
  filter = Expression$field_ref("CheckoutYear") == 2021,
  use_threads = FALSE
)
system.time(single_threaded_scan$ToTable())
#>    user  system elapsed 
#>  16.699   5.840  20.603
```

<sup>Created on 2022-12-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

binary-data-formats.qmd
djnavarro: good idea!

djnavarro:
> This looks to be very close to done to me. I have a couple last comments/questions below and then I think we can merge grinning

I think it's probably at a merge-able state 🙃. There are things that are still bothering me - as usual - but the current version feels like a readable, coherent, and self-contained block of content I think!

binary-data-formats.qmd
hadley: Nice, thanks!

hadley:
When I run 

```R
seattle_lib |>
  group_by(CheckoutYear) |>
  write_dataset(
    path = "data/seattle-library-checkouts",
    format = "parquet"
  )
```

I get "Invalid: In CSV column #7: Row #83240: CSV conversion error to null: invalid value '9781504752848'", because the type for `isbn` is imputed as `null` (since the first rows don't have it). What's the simplest way to avoid this problem?

stephhazlitt:
I am getting the same error (but on different row, which I think is just related to random order of array chunks being read):

`FileSystemDataset with 1 csv file
Error: Invalid: CSV parse error: Row #11742586: Expected 12 columns, got 2: Physical,Hori`

Could this be related to [ARROW- 18219](https://issues.apache.org/jira/browse/ARROW-18219)?

djnavarro:
Progress. It looks like Steph and I have subtly different versions of the csv file. When I read my local copy `read_csv_arrow()` infers that `isbn` has type string. Steph has a local copy from which it infers type null.

Really, this is just another argument for why nobody should be using csv files 😁

hadley:
You can also now use the csv I put in S3, just to make sure we're all working with exactly the same file.

stephhazlitt:
Curious. I sourced the file in the S3 bucket and now cannot reproduce the issue.

`open_dataset()` w/o specifying the schema infers ISBN as `null` but still works:

```
> seattle_csv <- open_dataset(
+   sources = "data/seattle-library-checkouts.csv",
+   format = "csv"
+ )
> 
> seattle_csv$schema
Schema
UsageClass: string
CheckoutType: string
MaterialType: string
CheckoutYear: int64
CheckoutMonth: int64
Checkouts: int64
Title: string
ISBN: null
Creator: string
Subjects: string
Publisher: string
PublicationYear: string
> 
> seattle_csv |> glimpse()
FileSystemDataset with 1 csv file
41,389,465 rows x 12 columns
$ UsageClass      <string> "Physical", "Physical", "Digital", "Physi…
$ CheckoutType    <string> "Horizon", "Horizon", "OverDrive", "Horiz…
$ MaterialType    <string> "BOOK", "BOOK", "EBOOK", "BOOK", "SOUNDDI…
$ CheckoutYear     <int64> 2016, 2016, 2016, 2016, 2016, 2016, 2016,…
$ CheckoutMonth    <int64> 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,…
$ Checkouts        <int64> 1, 1, 1, 1, 1, 1, 1, 1, 4, 1, 1, 2, 3, 2,…
$ Title           <string> "Super rich : a guide to having it all / …
$ ISBN              <null> NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, N…
$ Creator         <string> "Simmons, Russell", "Barclay, James, 1965…
$ Subjects        <string> "Self realization, Conduct of life, Attit…
$ Publisher       <string> "Gotham Books,", "Pyr,", "Random House, I…
$ PublicationYear <string> "c2011.", "2010.", "2015", "2005.", "c200…
```

`open_dataset()` specifiying the schema works:
```
> seattle_csv <- open_dataset(
+   sources = "data/seattle-library-checkouts.csv", 
+   format = "csv",
+   schema = schema(
+     field("UsageClass", string()),
+     field("CheckoutType", string()),
+     field("MaterialType", string()),
+     field("CheckoutYear", int64()),
+     field("CheckoutMonth", int64()),
+     field("Checkouts", int64()),
+     field("Title", string()),
+     field("ISBN", string()),
+     field("Creator", string()),
+     field("Subjects", string()),
+     field("Publisher", string()),
+     field("PublicationYear", string())
+   ),
+   skip_rows = 1
+ )
> 
> seattle_csv$schema
Schema
UsageClass: string
CheckoutType: string
MaterialType: string
CheckoutYear: int64
CheckoutMonth: int64
Checkouts: int64
Title: string
ISBN: string
Creator: string
Subjects: string
Publisher: string
PublicationYear: string
> 
> seattle_csv |> glimpse()
FileSystemDataset with 1 csv file
41,389,465 rows x 12 columns
$ UsageClass      <string> "Physical", "Physical", "Digital", "Physi…
$ CheckoutType    <string> "Horizon", "Horizon", "OverDrive", "Horiz…
$ MaterialType    <string> "BOOK", "BOOK", "EBOOK", "BOOK", "SOUNDDI…
$ CheckoutYear     <int64> 2016, 2016, 2016, 2016, 2016, 2016, 2016,…
$ CheckoutMonth    <int64> 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,…
$ Checkouts        <int64> 1, 1, 1, 1, 1, 1, 1, 1, 4, 1, 1, 2, 3, 2,…
$ Title           <string> "Super rich : a guide to having it all / …
$ ISBN            <string> "", "", "", "", "", "", "", "", "", "", "…
$ Creator         <string> "Simmons, Russell", "Barclay, James, 1965…
$ Subjects        <string> "Self realization, Conduct of life, Attit…
$ Publisher       <string> "Gotham Books,", "Pyr,", "Random House, I…
$ PublicationYear <string> "c2011.", "2010.", "2015", "2005.", "c200…
```
I think my local file, sourced manually from the website, that errored was incomplete as the overall size on disk registers smaller than my S3 download.

hadley:
@stephhazlitt and `write_dataset()` works too?

stephhazlitt:
For me, yes. This code from the chapter worked:

```
seattle_csv |>
  group_by(CheckoutYear) |>
  write_dataset(path = path, format = "parquet")

**Comment by hadley** [comment]
What versions of arrow are you using? I just got the latest version from CRAN and still see the problem.

**Comment by djnavarro** [comment]
If it does work, I'm wondering if the "right" thing to do is to:

1. quietly shuffle some of the rows in the distributed data file so that at least one non-null ISBN appears early enough so that arrow correctly infers the schema. My feeling is that r4ds isn't the right place to dig into those details.
2. open a ticket on the arrow jira (or github if we ever finally migrate): unless I've missed it in the docs, it seems to me that the arrow csv reader is missing something pretty fundamental, because the user can't specify the number of rows that arrow should inspect before inferring a column type. If it is possible, then the documentation needs updating because I can't even find it in the C++ docs, much less the R docs. 

Thoughts?

**Comment by djnavarro** [comment]
Yet another update, courtesy of @amoeba pointing me in the right direction on slack! 😁 

Okay, I haven't completely pinned it down but this workaround ensures that arrow infers the correct schema it by ensuring that the first block from the CSV file that it reads is large enough to contain one ISBN that looks like a string:

```r
ds <- open_dataset(csv_path, 
                   format = "csv", 
                   read_options = CsvReadOptions$create(block_size = 200000000))
```

If I pull it back to block_size = 100000000, I get int64 type because there's a book with a single ISBN earlier in the CSV. The first stringlike multi-ISBN entry is later. That feels a bit irritating as a solution, and I have a suspicion it may be a bug.

djnavarro:
In case we need to reference this later, here's a minimal(ish) reprex:

``` r
library(arrow, warn.conflicts = FALSE)

# on my machine, the magic number of bad rows needed to make 
# open_dataset break with the default settings in this example
# is 116506, but we can make it work with any number
num_bad_rows <- 100
csv_values <- c(
  '"a","b","c"',
  rep('"x",,"z"', num_bad_rows),
  rep('"x","y","z"')
)
csv_file <- tempfile()
writeLines(csv_values, csv_file)

too_small <- file.size(csv_file) - 1
nice_size <- file.size(csv_file)

options1 <- CsvReadOptions$create(block_size = too_small)
options2 <- CsvReadOptions$create(block_size = nice_size)

ds1 <- open_dataset(csv_file, format = "csv", read_options = options1)
ds2 <- open_dataset(csv_file, format = "csv", read_options = options2)

ds1 # wrong schema
#> FileSystemDataset with 1 csv file
#> a: string
#> b: null
#> c: string
ds2 # right schema
#> FileSystemDataset with 1 csv file
#> a: string
#> b: string
#> c: string
```

<sup>Created on 2022-12-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley:
@djnavarro is there a way to override the default type for just one column? If so, I could use that and then just mention in the text where to learn more.

djnavarro:
Agreed!

## Solution

I only worked out how to do it this morning, but you can override a subset of columns. It's not well documented though and buried in the R6 classes. The trick is to pass a schema for the columns you want to override to `CsvConvertOptions$create()` and then pass that as the `convert_options` argument:

``` r
library(arrow, warn.conflicts = FALSE)
library(dplyr, warn.conflicts = FALSE)

csv_path <- "data/seattle-library-checkouts.csv"
pq_path <- "data/seattle-library-checkouts"

# partial schema for ISBN column only
opts <- CsvConvertOptions$create(col_types = schema(ISBN = string()))

ds <- open_dataset(csv_path, format = "csv", convert_options = opts)
ds |>
  group_by(CheckoutYear) |>
  write_dataset(path = pq_path, format = "parquet")
```

<sup>Created on 2022-12-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

## Extra detail

Regarding the underlying issue itself, I dug a little further and oddly enough it's actually intended behaviour for which the logic hasn't bubbled up from the C++ docs to the R docs. Under the hood the Arrow C++ library has two CSV readers. The TableReader (in R, `CsvTableReader`) is used when you want to read the whole table into memory, and it bases the type inference on the whole column. `read_csv_arrow()` uses `CsvTableReader`. The Seattle library data is just small enough for me to load so I can do this:

``` r
arrow::read_csv_arrow(
  "data/seattle-library-checkouts.csv", 
  as_data_frame = FALSE
)
#> Table
#> 41389465 rows x 12 columns
#> $UsageClass <string>
#> $CheckoutType <string>
#> $MaterialType <string>
#> $CheckoutYear <int64>
#> $CheckoutMonth <int64>
#> $Checkouts <int64>
#> $Title <string>
#> $ISBN <string>
#> $Creator <string>
#> $Subjects <string>
#> $Publisher <string>
#> $PublicationYear <string>
```

<sup>Created on 2022-12-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

It correctly infers that ISBN is a string because it looks all the way to the end of the column before assigning a type. That's *actually* what I did originally while mucking about with the very first draft of the chapter and why I never encountered the problem. Oops.

However, there is also a [streaming reader](https://arrow.apache.org/docs/cpp/api/formats.html#_CPPv4N5arrow3csv15StreamingReaderE) which I'm pretty certain the Dataset API relies on, since (by design) Datasets don't read the whole file into memory at once as the expectation is that the file could be larger than memory. Instead, the file is split into blocks of fixed size, and each block is streamed when requested. Relevant part of the documentation:

> Type inference is done on the first block and types are frozen afterwards; to make sure the right data types are inferred, either set [ReadOptions::block_size](https://arrow.apache.org/docs/cpp/api/formats.html#structarrow_1_1csv_1_1_read_options_1a5c7b8530d3ad282a2e0736d8ab992208) to a large enough value, or use [ConvertOptions::column_types](https://arrow.apache.org/docs/cpp/api/formats.html#structarrow_1_1csv_1_1_convert_options_1ae27dbfff052b92e277deaae33b702b11) to set the desired data types explicitly.

So the intuitive answer I came up with earlier was essentially the first solution. The convert solution above is the second. It's a little awkward that - as far as I can tell - none of this has percolated up into a user friendly discussion in the R documentation, but hopefully I'll be able to add more of this over time

hadley:
Thanks for the detailed explanation! I think one thing that's currently missing from the arrow R package is a version of `open_dataset()` that mimics the `read_csv_arrow()` interface, i.e. is something that would be familiar to a tidyverse user. Maybe it could be `read_csv_lazy()` or similar?

stephhazlitt:
🎯 https://issues.apache.org/jira/browse/ARROW-18358

hadley:
Have you written anything that compares arrow to databases? i.e. when should you convert your csv to parquet and use arrow and when should you load it into a database?

djnavarro:
It's not something I've ever looked into. At some point I probably should!

hadley:
If I don't know and you don't know, I'll just encourage people to experiment and figure it out themselves 😄

arrow.qmd
```diff
@@ -0,0 +1,292 @@
...
+source("_common.R")
+status("polishing")
+```
+
+## Introduction
+
+CSV files are designed to be easily read by humans.
+They're a good interchange format because they're very simple and they can be read by every tool under the sun.
+But CSV files aren't very efficient: you have to do quite a lot of work to read the data into R.
+In this chapter, you'll learn about a powerful alternative: the [parquet format](https://parquet.apache.org/), an open standards-based format widely used by big data systems
```
mine-cetinkaya-rundel, line 19:
```suggestion
In this chapter, you'll learn about a powerful alternative: the [parquet format](https://parquet.apache.org/), an open standards-based format widely used by big data systems.
```

arrow.qmd
```diff
@@ -0,0 +1,292 @@
...
+
+We'll pair parquet files with [Apache Arrow](https://arrow.apache.org), a multi-language toolbox designed for efficient analysis and transport of large data sets.
+We'll use Apache Arrow via the the [arrow package](https://arrow.apache.org/docs/r/), which provides a dplyr backend allowing you to analyze larger-than-memory datasets using familiar dplyr syntax.
+As an additional benefit, arrow is extremely fast: you'll see some examples later in the chapter.
+
+Both arrow and dbplyr provide dplyr backends, so you might wonder when to use each.
+In many cases, the choice is made for you, as if the data is already in a database or in parquet files, you'll want to work with it as is.
```
mine-cetinkaya-rundel, line 26:
```suggestion
In many cases, the choice is made for you, as in the data is already in a database or in parquet files, and you'll want to work with it as is.
```

arrow.qmd
```diff
@@ -0,0 +1,292 @@
...
+But if you're starting with your own data (perhaps CSV files), you can either load into a database or convert to parquet.
```
mine-cetinkaya-rundel, line 27:
```suggestion
But if you're starting with your own data (perhaps CSV files), you can either load it into a database or convert it to parquet.
```

arrow.qmd
```diff
@@ -0,0 +1,292 @@
...
+In general, it's hard to know what will work best, so in the early stages of your analysis we'd encourage you to try both and pick the one that works the best for you.
+
+### Prerequisites
+
+In this chapter, we'll continue to use the tidyverse, particularly dplyr, but now we'll pair it with the arrow package which is designed specifically for working with large data.
```
mine-cetinkaya-rundel, line 32:
```suggestion
In this chapter, we'll continue to use the tidyverse, particularly dplyr, but we'll pair it with the arrow package which is designed specifically for working with large data.
```

arrow.qmd
```diff
@@ -0,0 +1,292 @@
...
+Later in the chapter, we'll also see some connections between arrow and duckdb, so we'll also need dbplyr and duckdb.
+
+```{r}
+library(dbplyr)
+library(duckdb)
+```
+
+## Getting the data
+
+We being by getting a dataset worthy of these tools: a data set of item checkouts from Seattle public libraries, available online at [data.seattle.gov/Community/Checkouts-by-Title/tmmm-ytt6](https://data.seattle.gov/Community/Checkouts-by-Title/tmmm-ytt6).
```
mine-cetinkaya-rundel, line 50:
```suggestion
We begin by getting a dataset worthy of these tools: a data set of item checkouts from Seattle public libraries, available online at [data.seattle.gov/Community/Checkouts-by-Title/tmmm-ytt6](https://data.seattle.gov/Community/Checkouts-by-Title/tmmm-ytt6).
```

arrow.qmd
```diff
@@ -0,0 +1,292 @@
...
+`open_dataset()` will scan a few thousand rows to figure out the structure of the data set.
+Then it records what it's found and stops; it will only read further rows as you specifically request them.
+This metadata is what we see if we print `seattle_csv`:
+
+```{r}
+seattle_csv
+```
+
+The first line in the output tell you that `seattle_csv` is stored locally on-disk as a single csv file; it will only be loaded into memory when necessary.
+The remainder of the output tell you the column type that arrow has imputed for each column.
```
mine-cetinkaya-rundel, line 96:
```suggestion
The remainder of the output tells you the column type that arrow has imputed for each column.
```

arrow.qmd
```diff
@@ -0,0 +1,292 @@
...
+What happens when this code is run?
+`open_dataset()` will scan a few thousand rows to figure out the structure of the data set.
+Then it records what it's found and stops; it will only read further rows as you specifically request them.
+This metadata is what we see if we print `seattle_csv`:
+
+```{r}
+seattle_csv
+```
+
+The first line in the output tell you that `seattle_csv` is stored locally on-disk as a single csv file; it will only be loaded into memory when necessary.
```
mine-cetinkaya-rundel, line 95:
```suggestion
The first line in the output tells you that `seattle_csv` is stored locally on-disk as a single CSV file; it will only be loaded into memory when necessary.
```

arrow.qmd
```diff
@@ -0,0 +1,292 @@
...
+```{r show-parquet-files}
+tibble(
+  files = list.files(pq_path, recursive = TRUE),
+  size_MB = file.size(file.path(pq_path, files)) / 1024^2
+)
+```
+
+Our single 9GB CSV file has been rewritten into 18 parquet files.
+The file names use a "self-describing" convention used by the [Apache Hive](https://hive.apache.org) project.
+Hive-style partitions name folders with a "key=value" convention, so as you might guess the `CheckoutYear=2005` directory contains all the data where `CheckoutYear` is 2005.
```
mine-cetinkaya-rundel, line 190:
```suggestion
Hive-style partitions name folders with a "key=value" convention, so as you might guess, the `CheckoutYear=2005` directory contains all the data where `CheckoutYear` is 2005.
```

arrow.qmd
```diff
@@ -0,0 +1,292 @@
...
+  collect() |> 
+  system.time()
+```
+
+The \~100x speedup in performance is attributable to two factors: the multi-file partitioning, and the format of individual files:
+
+-   Partitioning improves performance because this query uses `CheckoutYear == 2021` to filter the data, and arrow is smart enough to recognize that it only needs to read 1 of the 18 parquet files.
+-   The parquet format improves performance by storing data in a binary format that can be read more directly into memory. The column-wise format and rich metadata means that arrow only needs to read the four columns actually used in the query (`CheckoutYear`, `MaterialType`, `CheckoutMonth`, and `Checkouts`).
+
+This massive difference is performance is why that whenever you're working with large CSVs, you should always convert to parquet first!
```
mine-cetinkaya-rundel, line 266:
```suggestion
This massive difference in performance is why whenever you're working with large CSVs, you should always convert to parquet first!
```

mine-cetinkaya-rundel (approved):
My comments are just a few typos I caught while I read through the chapter

hadley:
@djnavarro Thanks for all your hard work on this! Feel free to do another PR if anything else comes mind (or your see typos that I've introduced), but overall the chapter feels like a really useful additional to the book and I really appreciate the time you put into it.

djnavarro:
Sure! Happy to take a look and see if there's anything I can add. Turns out I have a lot of free time these days 🫠

PR #1138, opened 2022-11-17
Use .by instead of group_by()

hadley:
Minimal set of updates. I'm not seriously considering this change for this edition but I wanted to see what it would look like.

cc @mine-cetinkaya-rundel, @davisvaughan

### DIFF
--- a/base-R.qmd
+++ b/base-R.qmd
@@ -422,8 +422,10 @@ Another important member of the apply family is `tapply()` which computes a sing
 
 ```{r}
 diamonds |> 
-  group_by(cut) |> 
-  summarise(price = mean(price))
+  summarise(
+    price = mean(price),
+    .by = cut
+  )
 
 tapply(diamonds$price, diamonds$cut, mean)
 ```
--- a/base-R.qmd
+++ b/base-R.qmd
@@ -423,8 +423,8 @@ Another important member of the apply family is `tapply()` which computes a sing
 ```{r}
 diamonds |> 
   summarise(
+    .by = cut,
     price = mean(price),
-    .by = cut
   )
 
 tapply(diamonds$price, diamonds$cut, mean)
--- a/data-tidy.qmd
+++ b/data-tidy.qmd
@@ -569,8 +569,10 @@ Since you don't know how to work with this sort of data yet, you'll want to foll
 
 ```{r}
 df |> 
-  group_by(id, name) |> 
-  summarize(n = n(), .groups = "drop") |> 
+  summarize(
+    n = n(),
+    .by = c(id, name)
+  ) |> 
   filter(n > 1L) 
 ```
 
@@ -678,10 +680,10 @@ For example, if you're interested in just the total number of missing values in 
 
 ```{r}
 cms_patient_experience |> 
-  group_by(org_pac_id) |> 
   summarize(
     n_miss = sum(is.na(prf_rate)),
     n = n(),
+    .by = org_pac_id
   )
 ```
 
--- a/data-tidy.qmd
+++ b/data-tidy.qmd
@@ -570,8 +570,8 @@ Since you don't know how to work with this sort of data yet, you'll want to foll
 ```{r}
 df |> 
   summarize(
+    .by = c(id, name),
     n = n(),
-    .by = c(id, name)
   ) |> 
   filter(n > 1L) 
 ```
@@ -681,9 +681,9 @@ For example, if you're interested in just the total number of missing values in 
 ```{r}
 cms_patient_experience |> 
   summarize(
+    .by = org_pac_id,
     n_miss = sum(is.na(prf_rate)),
     n = n(),
-    .by = org_pac_id
   )
 ```
 
--- a/data-transform.qmd
+++ b/data-transform.qmd
@@ -74,9 +74,9 @@ That makes it possible to get a sense of the following code even though you have
 
 flights |>
   filter(dest == "IAH") |> 
-  group_by(year, month, day) |> 
   summarize(
-    arr_delay = mean(arr_delay, na.rm = TRUE)
+    arr_delay = mean(arr_delay, na.rm = TRUE),
+    .by = c(year, month, day)
   )
 ```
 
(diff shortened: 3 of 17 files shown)


data-tidy.qmd
```diff
@@ -569,8 +569,10 @@ Since you don't know how to work with this sort of data yet, you'll want to foll
 
 ```{r}
 df |> 
-  group_by(id, name) |> 
-  summarize(n = n(), .groups = "drop") |> 
+  summarize(
+    .by = c(id, name),
+    n = n(),
+  ) |>
```
DavisVaughan, line 575: This is one where I'd just write

```r
summarise(n = n(), .by = c(id, name))
```

I'm not sure we need to commit to up-front-`.by` religiously.

I feel it is most useful if you have multiple expressions (or one multiline expression) in your summarise/mutate call, and you want to pull it up front to ensure it doesn't get "lost" in the details

Otherwise my brain seems to want to type out the expression first, and the `.by` second, and I'm not sure how much I want to fight that

data-tidy.qmd
```diff
@@ -569,8 +569,10 @@ Since you don't know how to work with this sort of data yet, you'll want to foll
 
 ```{r}
 df |> 
-  group_by(id, name) |> 
-  summarize(n = n(), .groups = "drop") |> 
+  summarize(
+    .by = c(id, name),
+    n = n(),
+  ) |> 
   filter(n > 1L)
```
DavisVaughan, line 576: In terms of teaching, is it easier to explain that you go from:

```r
df %>% summarise(x = mean(x))
```

to:

```r
df %>% summarise(x = mean(x), .by = c(g1, g2))
```

or:

```r
df %>% summarise(.by = c(g1, g2), x = mean(x))
```

?

The latter feels somewhat unnatural for beginners who aren't used to the idea that the placement of named arguments is arbitrary. They are also likely using simple `summarise()` expressions where it is unlikely that the `.by` call will get "lost"

data-transform.qmd
```diff
@@ -473,76 +467,73 @@ For example, the following code finds the most delayed flight to each destinatio
...
-daily |> 
-  ungroup() |>
-  summarize(
-    delay = mean(dep_delay, na.rm = TRUE), 
-    flights = n()
+flights |> 
+  mutate(
+    .by = year:day,
+    rank = min_rank(dep_delay),
+    .after = dep_delay,
```
DavisVaughan, line 527: particularly odd to have `.by` separated from `.after` - both optional args

data-transform.qmd
```diff
@@ -625,10 +616,10 @@ The following code uses data from the **Lahman** package to compare what proport
 
 ```{r}
 batters <- Lahman::Batting |> 
-  group_by(playerID) |> 
   summarize(
     perf = sum(H, na.rm = TRUE) / sum(AB, na.rm = TRUE),
-    n = sum(AB, na.rm = TRUE)
+    n = sum(AB, na.rm = TRUE),
+    by = playerID
```
DavisVaughan, line 622: uncaught typo of `by` vs `.by`

i dont think we can guard against this though.

data-tidy.qmd
mine-cetinkaya-rundel: I guess my comment above is related to this. `.by` upfront feels weird.

data-tidy.qmd
mine-cetinkaya-rundel: Agreed

data-transform.qmd
```diff
@@ -473,76 +467,73 @@ For example, the following code finds the most delayed flight to each destinatio
...
 
 ```{r}
-daily <- flights |>  
-  group_by(year, month, day)
-daily
+flights |> 
+  filter(
+    .by = dest,
+    arr_delay == max(arr_delay, na.rm = TRUE),
+  )
```
mine-cetinkaya-rundel, line 490: How does one read this?

"Take the flights data frame and filter it, by each destination, for observations by where arrival delay is equal to the maximum arrival delay for that destination."

Since we didn't do `group_by()` first, I felt like "maximum arrival delay *for that destination*" is not immediately clear.

I also think `.by` should be the second argument but these are separate concerns.

data-transform.qmd
mine-cetinkaya-rundel: +1

workflow-pipes.qmd
```diff
@@ -35,32 +35,33 @@ For example, the last chapter finished with a moderately complex pipe:
...
+    .by = tailnum,
     delay = mean(arr_delay, na.rm = TRUE),
     n = n()
-  )
+  ) |> 
+  filter(n > 5)
 ```
 
-Even though this pipe has four steps, it's easy to skim because the verbs come at the start of each line: start with the `flights` data, then filter, then group, then summarize.
+Even though this pipe has thre steps, it's easy to skim because the verbs come at the start of each line: start with the `flights` data, then filter, then group, then summarize.
```
mine-cetinkaya-rundel, line 46:
```suggestion
Even though this pipe has three steps, it's easy to skim because the verbs come at the start of each line: start with the `flights` data, then filter, then group, then summarize.
```

workflow-pipes.qmd
```diff
@@ -35,32 +35,33 @@ For example, the last chapter finished with a moderately complex pipe:
...
 What would happen if we didn't have the pipe?
 We could nest each function call inside the previous call:
 
 ```{r}
 #| eval: false
 
-summarise(
-  group_by(
+filter(
+  summarise(
```
mine-cetinkaya-rundel, line 55: No 🇬🇧, resist the urge!

mine-cetinkaya-rundel (commented):
Something I hadn't thought about until reviewing this -- in articulating the pipeline ("reading out loud"), with `.by` instead of `group_by()` the word "group" is completely lost. Is that intentional? It might just be fine, but people won't know to say "group by" when they see `.by` and "by" on its own isn't as informative as "group by". 🤔 Is it a terrible idea for it to be `.group_by`?

base-R.qmd
```diff
@@ -422,8 +422,10 @@ Another important member of the apply family is `tapply()` which computes a sing
 
 ```{r}
 diamonds |> 
-  group_by(cut) |> 
-  summarise(price = mean(price))
+  summarise(
+    .by = cut,
+    price = mean(price),
+  )
```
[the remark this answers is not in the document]
DavisVaughan: > calculate the mean price by cut.

Just noting that this feels like a natural way to state this and you never say "group by" here either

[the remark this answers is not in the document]
mine-cetinkaya-rundel: Agreed. My concern was that people won't even have the word "group" as part of their vocabulary. This may not be a huge concern, just something I realized getting lost in the articulation compared to the `group_by() |> summarize()` flow.

hadley:
Closing since we don't plan on actually doing this, but the discussion was super helpful 😄

PR #1148, opened 2022-11-23
Delete redundant words Running R Code section

aephidayatuloh:
At the and of Running R Code section the "we'll use" are written 2 times. So I deleted one of them.


mine-cetinkaya-rundel:
Thank you!

PR #1151, opened 2022-11-24
ggsave's first attribute is filename, not path

werkstattcodes:
I think it's potentially somewhat misleading to use for ggsave "path" instead of "filename" in the example below.  I might be wrong, but I think what happens in the proposed

` ggsave(path, plot, width = 6, height = 6)`

is actually 

`ggsave(filename=path, path=NULL, plot=plot, width=6, height=6). `

It might be clearer for the reader, to use filename instead of path:

`ggsave(filename=filename, plot=plot, width=6, height=6)`

At least I found it somewhat confusing when reading the text. But feel free to ignore.

### DIFF
--- a/functions.qmd
+++ b/functions.qmd
@@ -798,7 +798,7 @@ foo(cyl)
 ```
 
 As with data frame functions, it can be useful to make your plotting functions tightly coupled to a specific dataset, or even a specific variable.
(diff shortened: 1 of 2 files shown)


hadley:
I think the way that `ggsave()` is constructed is confusing, but in most cases, you will actually pass a path to the first argument.

PR #1156, opened 2022-11-28
Small typo and grammar edit

mmhamdy:
Edits in the data visualization chapter.


hadley:
Thanks for the fix, but we're in the middle of a big re-org of this (and related) chapters, so I think it's easier to not merge it right now to avoid the potential for merge conflicts.

PR #1167, opened 2022-11-30
Update logicals - case_when - reorder arguments

p0bs:
OK, so this is my first PR, but I think the order of your original `case_when` in this section creates an issue, in that nothing is ever deemed "very early". For example:

```r
> flights |> 
+     mutate(
+         status = case_when(
+             is.na(arr_delay)      ~ "cancelled",
+             arr_delay > 60        ~ "very late",
+             arr_delay > 15        ~ "late",
+             abs(arr_delay) <= 15  ~ "on time",
+             arr_delay < -15       ~ "early",
+             arr_delay < -30       ~ "very early",
+         ),
+         .keep = "used"
+     ) |> slice(54)
# A tibble: 1 × 2
  arr_delay status
      <dbl> <chr> 
1       -33 early 
```

... whereas 

```r
> flights |> 
+     mutate(
+         status = case_when(
+             is.na(arr_delay)      ~ "cancelled",
+             arr_delay > 60        ~ "very late",
+             arr_delay > 15        ~ "late",
+             abs(arr_delay) <= 15  ~ "on time",
+             arr_delay < -30       ~ "very early",
+             arr_delay < -15       ~ "early"
+         ),
+         .keep = "used"
+     ) |> slice(54)
# A tibble: 1 × 2
  arr_delay status    
      <dbl> <chr>     
1       -33 very early
```


hadley:
Oops, thanks for spotting this! I think it would be slightly better to reorder the whole thing so there’s a consistent run from smallest to largest numbers. And maybe it’s worth even keeping the original code in as a cautionary tale.

PR #1168, opened 2022-12-01
Patch 2

mcsnowface:
Found some typos and also have one small suggestion (see comment).

### DIFF
--- a/numbers.qmd
+++ b/numbers.qmd
@@ -665,7 +665,7 @@ For these reasons, the mode tends not to be used by statisticians and there's no
 
 What if you're interested in locations other than the center?
 `min()` and `max()` will give you the largest and smallest values.
-Another powerful tool is `quantile()` which is a generalization of the median: `quantile(x, 0.25)` will find the value of `x` that is greater than 25% of the values, `quantile(x, 0.5)` is equivalent to the median, and `quantile(x, 0.95)` will find a value that's greater than 95% of the values.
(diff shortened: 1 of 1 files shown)


numbers.qmd
```diff
@@ -273,12 +273,12 @@ starting <- 100
 interest <- 1.05
 
 money <- tibble(
-  year = 2000 + 1:50,
```
mcsnowface, line 276: In this section the name "year" has been used in two different ways: a) the number of years interest is compounding (1:50), and b) in the form of the calendar year staring from 2000. For a) see here: "That gives a formula like money = starting * interest ^ year" ; and for b) the variable name "year" in the tibble. I think it would be more clear to only apply one meaning to the name "year" in this example.

numbers.qmd
```diff
@@ -293,12 +293,12 @@ ggplot(money, aes(year, money)) +
   scale_y_log10()
 ```
 
-This a straight line because a little algebra reveals that `log(money) = log(starting) + n * log(interest)`, which matches the pattern for a line, `y = m * x + b`.
+This a straight line because a little algebra reveals that `log10(money) = log10(interest) * year + log10(starting)`, which matches the pattern for a line, `y = m * x + b`.
```
mcsnowface, line 296: To keep it as clear as possible I would add log10 here as well as swap the parts of the formula so that it actually matches the pattern of a line, i.e. y = mx +b.

hadley:
Thanks!

PR #1178, opened 2022-12-08
Update functions.qmd

mcsnowface:
Did some minor edits and changes. Also added two comments.

### DIFF
--- a/functions.qmd
+++ b/functions.qmd
@@ -75,7 +75,7 @@ Preventing this type of mistake of is one very good reason to learn how to write
 ### Writing a function
 
 To write a function you need to first analyse your repeated code to figure what parts are constant and what parts vary.
-If we take the code above and pull it outside of `mutate()` it's a little easier to see the pattern because each repetition is now one line:
+If we take the code above and pull it outside of `mutate()`, it's a little easier to see the pattern because each repetition is now one line:
 
 ```{r}
 #| eval: false
@@ -99,11 +99,11 @@ To turn this into a function you need three things:
     Here we'll use `rescale01` because this function rescales a vector to lie between 0 and 1.
 
 2.  The **arguments**.
-    The arguments are things that vary across calls and our analysis above tells us that have just one.
+    The arguments are things that vary across calls and our analysis above tells us that we have just one.
     We'll call it `x` because this is the conventional name for a numeric vector.
 
 3.  The **body**.
-    The body is the code that repeated across all the calls.
+    The body is the code that's repeated across all the calls.
 
 Then you create a function by following the template:
 
@@ -143,7 +143,7 @@ df |> mutate(
 
 ### Improving our function
 
-You might notice `rescale01()` function does some unnecessary work --- instead of computing `min()` twice and `max()` once we could instead compute both the minimum and maximum in one step with `range()`:
+You might notice that the `rescale01()` function does some unnecessary work --- instead of computing `min()` twice and `max()` once we could instead compute both the minimum and maximum in one step with `range()`:
 
 ```{r}
 rescale01 <- function(x) {
@@ -166,6 +166,7 @@ rescale01 <- function(x) {
   rng <- range(x, na.rm = TRUE, finite = TRUE)
   (x - rng[1]) / (rng[2] - rng[1])
 }
+
 rescale01(x)
 ```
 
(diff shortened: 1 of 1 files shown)


functions.qmd
```diff
@@ -275,22 +280,24 @@ haversine <- function(long1, lat1, long2, lat2, round = 3) {
 
 ### Summary functions
 
-Another important family of vector functions is summary functions, functions that return a single value for use in `summarize()`.
+Another important family of vector functions are summary functions, functions that return a single value for use in `summarize()`.
```
mine-cetinkaya-rundel, line 283:
```suggestion
Another important family of vector functions is summary functions, functions that return a single value for use in `summarize()`.
```

functions.qmd
```diff
@@ -166,26 +166,27 @@ rescale01 <- function(x) {
...
 ```
 
 These changes illustrate an important benefit of functions: because we've moved the repeated code into a function, we only need to make the change in one place.
 
 ### Mutate functions
 
-Now you've got the basic idea of functions, lets take a look a whole bunch of examples.
-We'll start by looking at "mutate" functions, functions that work well like `mutate()` and `filter()` because they return an output the same length as the input.
+Now you've got the basic idea of functions, let's take a look at a whole bunch of examples.
+We'll start by looking at "mutate" functions, i.e. functions that work like `mutate()` and `filter()` because they return an output of the same length as the input.
```
mcsnowface, line 178: Maybe I am not understanding it correctly (it's late already). How do `filter()` and `mutate()` return an output of the same length as the input? Mutate can add new columns to the tibble and filter often reduces the amount of rows.

functions.qmd
```diff
@@ -717,14 +729,15 @@ hex_plot <- function(df, x, y, z, bins = 20, fun = "mean") {
...
+
 diamonds |> hex_plot(carat, price, depth)
 ```
 
 ### Combining with dplyr
 
 Some of the most useful helpers combine a dash of dplyr with ggplot2.
 For example, if you might want to do a vertical bar chart where you automatically sort the bars in frequency order using `fct_infreq()`.
-Since the bar chart is vertical, we also need to reverse the usual order to get the highest values at the top:
+Since the bar chart is vertical, we also need to reverse the usual order to get the highest values at the top (also note the `:=` operator, which allows you to inject names with glue syntax on the left-hand side of ⁠`:=`;⁠ type: ?\`:=\` for more details):
```
mcsnowface, line 741: This is just a suggestion but the ":=" is new (I haven't seen it myself either before). So perhaps pointing it out may be a good idea? My edit here is just a suggestion. Doesn't have to be that way.

functions.qmd
```diff
@@ -733,10 +746,11 @@ sorted_bars <- function(df, var) {
     ggplot(aes(y = {{ var }})) + 
     geom_bar()
 }
+
 diamonds |> sorted_bars(cut)
 ```
 
-Or you could maybe you want to make it easy to draw a bar plot just for a subset of the data:
+Or you maybe you want to make it easy to draw a bar plot just for a subset of the data:
```
mine-cetinkaya-rundel, line 753:
```suggestion
Or maybe you want to make it easy to draw a bar plot just for a subset of the data:
```

functions.qmd
```diff
@@ -166,26 +166,27 @@ rescale01 <- function(x) {
   rng <- range(x, na.rm = TRUE, finite = TRUE)
   (x - rng[1]) / (rng[2] - rng[1])
 }
+
```
mcsnowface, line 169: Throughout I tried to format the code so that it looks the same, i.e. adding a space whenever a new function has been defined. Please ignore if it this is unnecessary. It's just something I noticed.

mine-cetinkaya-rundel (commented):
A few edits on the changes, but I'll leave the comments for @hadley on this one.

functions.qmd
hadley: I think this should be "work well _inside_ of `mutate()` and `filter()`"

hadley, line 178:
```suggestion
We'll start by looking at "mutate" functions, i.e. functions that work well inside of `mutate()` and `filter()` because they return an output of the same length as the input.
```

functions.qmd
hadley: I'll flesh this out a bit more after merging.

hadley:
Thanks for these fixes!

PR #1179, opened 2022-12-10
Update iteration.qmd

mcsnowface:
Just some minor edits and one comment. I hope that's useful.


iteration.qmd
```diff
@@ -486,7 +486,6 @@ We'll then discuss how you can handle situations of increasing heterogeneity, wh
 ### Listing files in a directory
 
 As the name suggests, `list.files()` lists the files in a directory.
-TO CONSIDER: why not use it via the more obvious name `list.files()`?
```
mcsnowface, line 489: Maybe I misread something but the first sentence says 'list.files()'  and the second sentence says "[...] via the more obvious name `list.files()`. Both sentences contain the same function, i.e. `list.files()`.  Please ignore if I didn't get that right.

hadley:
Thanks! Your suggestions are super useful 😄

PR #1186, opened 2022-12-16
Fixed typos

KittJonathan:


data-transform.qmd
```diff
@@ -535,7 +535,7 @@ daily_flights <- daily |>
   )
 ```
 
-If you're happy with this behavior, you can explicitly request it in order to suppress the message:
+If you're not happy with this behavior, you can explicitly request it in order to suppress the message:
```
hadley, line 538: I don't this should be negated; specifying `drop_last` means that you're happy with the default behaviour, you just don't want the message.

KittJonathan: Makes sense :-)

mine-cetinkaya-rundel, line 538:
```suggestion
If you're happy with this behavior, you can explicitly request it in order to suppress the message:
```

mine-cetinkaya-rundel:
Thank you!

PR #1232, opened 2023-01-20
Text insertion and arguments update/alignment

stephenbalogun:
In EDA, line 379, ggplot2 prints a warning that the `size` argument has been deprecated and should be replaced with `linewidth`. I did the same replacement in line 407.

In communication, I added "to" in line 844 to read "It is generally more useful if you want to *expand* the limits..."

In line 1101, I changed the legend.position argument to "top" to align with what was written in line 1053.


hadley:
Thanks!

PR #1233, opened 2023-01-20
Update old function name

gl-eb:
The function `append_csv()` is nowhere to be found the rest of the book. I assume from the context that `append_csv` is an earlier name of what is now `append_file`


hadley:
Thanks!

PR #1234, opened 2023-01-21
Changes to logicals.qmd

stephenbalogun:
The condition mentioned was not met. I suggest that the statement be modified to read:
`case_when(x > 0 ~ "+ve", x > 2 ~ "big")`


mine-cetinkaya-rundel:
Thank you!

PR #1237, opened 2023-01-22
Update regexps

stephenbalogun:
1. Line 255: changed the function `separate_wider_location()` to what I believe is the appropriate one in the package - `separate_wider_position()`
2. Line 256: changed "because the operates on ..." to "because *they operate* on..."
3. Line 380: changed "`$` should matches..." to "`$` should *match*..."
4. Line 390: inserted *useful* in the sentence "this is particularly when using..."
5. Line 499: inserted *as* in the sentence "as well overriding operator precedence..."
6. Line 597: changed "that can use to control..." to "that can *be used* to control..."
7. Line 683: changed "checking you work..." to "checking *your* work..."
8. Line 684: deleted the double word "create create"
9. Line 865: changed the function `separate_with_regex()` to what I believe is the correct version in the package `separate_wider_regex()`
10. Line 868: reorder the functions `separate_delim_longer()` and `separate_delim_wider()` to the right ones contained in the package `separate_longer_delim()` and `separate_wider_delim()`


hadley:
Thanks for the careful read through!

PR #1244, opened 2023-01-23
Minor corrections

stephenbalogun:
## joins.qmd

1. Line 38: changed "... with on each table" to "...*within* each table".
2. Line 141: inserted *we* in "...if have some way to describe them to others"
3. Line 197: six joins were mentioned but 5 were listed. I believe the sixth is `full_join()`
4. Line 324: changed "...we can find rows that as missing from..." to "we can find rows that *are* missing from..."

## spreadsheets.qmd

1. Line 407: I believe the *writexl* package was intended and not *readxl* since  the line is referencing a package for writing Excel


hadley:
Thanks!

BTW you might find it a bit easier if you do PRs from a branch, and there's no need to explain simple stuff (e.g. typo fixes) that's obvious from looking at the diffs.

PR #1259, opened 2023-01-31
changed and to or, which is the technically true description I believe

ercan7:
not super important but technically and should be or. It turns out that both variables have NAs but the same error would pop up even if only one was missing. Also or doesn't exclude "both".


hadley:
"And" sounds better to my ear; I don't think there's a particularly clear way to express the precise idea in English.

PR #1260, opened 2023-01-31
direct reader to the book's vision & usage of DuckDB

wibeasley:


hadley:
Thanks! I tackled this slightly differently.

PR #1266, opened 2023-02-08
line 325

yingxingwu:
str_view(dot) resulted in Error of missing argument. 
suggested change: writeLines(dot) will return the correct output


yingxingwu:
updated my stingr package to 1.5.0, then the code worked. Sorry for the false request.

PR #1285, opened 2023-02-15
Minor Correction in 6.5 Summary Section

MatthiasLiew:
vignette(pivot, package = "tidyr") should be changed to vignette("pivot", package = "tidyr")

The former code produces an "object not found" error as vignette does not quote its arguments.


mine-cetinkaya-rundel:
Thank you!

MatthiasLiew:
You're welcome!

PR #1286, opened 2023-02-15
Fixed typos

KittJonathan:
- removed capital S in "so"
- changed text for plots shown on top of each other rather than next to each other (left-right -> first-second)


mine-cetinkaya-rundel:
I fixed the plot layout to be side-by-side and implemented the other fix in a separate commit, thank you for catching these!

PR #1298, opened 2023-02-23
a missing word inserted

ercan7:


ercan7:
In the book, it says "whenever you plot a mean (or other summary) vs. group size, you’ll see that the variation decreases as the sample size increases". This is a result of the LLN because according to the LLN, as the sample increases the sample mean converges to the population mean and therefore the variation is gonna be less. On the other hand, the CLT is about the distribution of the sample means, not the variation in sample means.

data-transform.qmd
```diff
@@ -11,7 +11,7 @@ status("complete")
 
 Visualisation is an important tool for generating insight, but it's rare that you get the data in exactly the right form you need for it.
 Often you'll need to create some new variables or summaries to see the most important patterns, or maybe you just want to rename the variables or reorder the observations to make the data a little easier to work with.
-You'll learn how to do all that (and more!) in this chapter, which will introduce you to data transformation using the **dplyr** package and a new dataset on flights that departed New York City in 2013.
+You'll learn how to do all that (and more!) in this chapter, which will introduce you to data transformation using the **dplyr** package and a new dataset on flights that departed from New York City in 2013.
```
hadley, line 14: I don't think from is needed here.

data-transform.qmd
```diff
@@ -699,9 +699,9 @@ ggplot(delays, aes(x = n, y = delay)) +
 ```
 
 Not surprisingly, there is much greater variation in the average delay when there are few flights for a given plane.
-The shape of this plot is very characteristic: whenever you plot a mean (or other summary) vs. group size, you'll see that the variation decreases as the sample size increases[^data-transform-4].
+The shape of this plot is very characteristic: whenever you plot a mean (or other summary statistics) vs. group size, you'll see that the variation decreases as the sample size increases[^data-transform-4].
```
hadley, line 702: I think using summary here is fine.

data-transform.qmd
```diff
@@ -699,9 +699,9 @@ ggplot(delays, aes(x = n, y = delay)) +
...
 
-[^data-transform-4]: \*cough\* the central limit theorem \*cough\*.
+[^data-transform-4]: \*cough\* the law of large numbers \*cough\*.
```
hadley, line 704: Good catch!

data-transform.qmd
mine-cetinkaya-rundel, line 14:
```suggestion
You'll learn how to do all that (and more!) in this chapter, which will introduce you to data transformation using the **dplyr** package and a new dataset on flights that departed New York City in 2013.
```

mine-cetinkaya-rundel:
Thank you!

PR #1300, opened 2023-02-24
Fixed error in text

KittJonathan:
Replaced "After reading the first two parts" with "After reading the first part" : I think this comes from reordering the chapters in the 2e ?


mine-cetinkaya-rundel:
Thank you!

PR #1307, opened 2023-02-27
Fixed typos

KittJonathan:
Fixed typos in figure explanation : top/middle/bottom rather than left/middle/right, due to the display of figures in the text.


hadley:
We might need to fix the figure positions, rather than their labels.

hadley:
Nope, you're right.

PR #1314, opened 2023-02-28
Fix typos and UK spellings

penelopeysm:
Hello! I've only read through a bit of the book, but here's a PR fixing a few typos and UK spellings which I've noticed so far.

Separately, I didn't edit these lines, but there's some duplication with the `alpha` aesthetic being mentioned twice:

https://github.com/hadley/r4ds/blob/e3cb55634424809d1d5a26c612fe13fbf9dd7d59/layers.qmd#L97

https://github.com/hadley/r4ds/blob/e3cb55634424809d1d5a26c612fe13fbf9dd7d59/layers.qmd#L128

### DIFF
--- a/arrow.qmd
+++ b/arrow.qmd
@@ -142,8 +142,8 @@ This means that:
     In contrast, parquet files store data in a way that records the type along with the data.
 
 -   Parquet files are "column-oriented".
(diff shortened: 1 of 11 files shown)


mine-cetinkaya-rundel:
Thank you!

PR #1330, opened 2023-03-01
Review edits

mine-cetinkaya-rundel:
For intro and visualize and transform in whole game.


data-transform.qmd
```diff
@@ -665,107 +665,53 @@ As you can see, when you summarize an ungrouped data frame, you get a single row
...
   summarize(
-    perf = sum(H, na.rm = TRUE) / sum(AB, na.rm = TRUE),
+    performance = sum(H, na.rm = TRUE) / sum(AB, na.rm = TRUE),
     n = sum(AB, na.rm = TRUE)
   )
 batters
 ```
 
-When we plot the skill of the batter (measured by the batting average, `ba`) against the number of opportunities to hit the ball (measured by at bat, `ab`), you see two patterns:
+When we plot the skill of the batter (measured by the batting average, `performance`) against the number of opportunities to hit the ball (measured by at bat, `n`), you see two patterns:
```
hadley, line 681:
```suggestion
When we plot the skill of the batter (measured by the batting average, `performance`) against the number of opportunities to hit the ball (measured by times at bat, `n`), you see two patterns:
```

data-transform.qmd
```diff
@@ -665,107 +665,53 @@ As you can see, when you summarize an ungrouped data frame, you get a single row
 
 Whenever you do any aggregation, it's always a good idea to include a count (`n()`).
 That way, you can ensure that you're not drawing conclusions based on very small amounts of data.
-For example, let's look at the planes (identified by their tail number) that have the highest average delays:
```
hadley, line 668: We need to link to this section from https://r4ds.hadley.nz/numbers.html#numeric-summaries

data-visualize.qmd
```diff
@@ -203,27 +207,19 @@ Before we add more layers to this plot, let's pause for a moment and review the
...
 Like R, ggplot2 subscribes to the philosophy that missing values should never silently go missing.
 This type of warning is probably one of the most common types of warnings you will see when working with real data -- missing values are a very common issue and you'll learn more about them throughout the book, particularly in @sec-missing-values.
 For the remaining plots in this chapter we will suppress this warning so it's not printed alongside every single plot we make.
 
 ### Adding aesthetics and layers
 
-Scatterplots are useful for displaying the relationship between two variables, but it's always a good idea to be skeptical of any apparent relationship between two variables and ask if there may be other variables that explain or change the nature of this apparent relationship.
+Scatterplots are useful for displaying the relationship between two numerical variables, but it's always a good idea to be skeptical of any apparent relationship between two variables and ask if there may be other variables that explain or change the nature of this apparent relationship.
```
hadley, line 217:
```suggestion
Scatterplots are useful for displaying the relationship between two numeric variables, but it's always a good idea to be skeptical of any apparent relationship between two variables and ask if there may be other variables that explain or change the nature of this apparent relationship.
```

data-visualize.qmd
```diff
@@ -243,13 +239,14 @@ ggplot(
   geom_point()
 ```
 
-When a variable is mapped to an aesthetic, ggplot2 will automatically assign a unique value of the aesthetic (here a unique color) to each unique level of the variable (each of the three species), a process known as **scaling**.
+When a categorical variable is mapped to an aesthetic element, ggplot2 will automatically assign a unique value of the aesthetic (here a unique color) to each unique level of the variable (each of the three species), a process known as **scaling**.
```
hadley, line 242:
```suggestion
When a categorical variable is mapped to an aesthetic, ggplot2 will automatically assign a unique value of the aesthetic (here a unique color) to each unique level of the variable (each of the three species), a process known as **scaling**.
```

We don't appear to use element anywhere else.

data-visualize.qmd
hadley: Oh you use numerical everywhere? That now looks/sounds weird to me.

data-visualize.qmd
```diff
@@ -651,36 +646,32 @@ ggplot(penguins, aes(x = species, y = body_mass_g)) +
   geom_boxplot()
 ```
 
-Alternatively, we can make frequency polygons with `geom_freqpoly()`.
-`geom_freqpoly()` performs the same calculation as `geom_histogram()`, but instead of displaying the counts with bars, it uses lines instead.
-It's much easier to understand overlapping lines than bars of `geom_histogram()`.
-There are a few challenges with this type of plot, which we will come back to in @sec-cat-num on exploring the covariation between a categorical and a numerical variable.
+Alternatively, we can make density plots with `geom_density()`.
```
hadley, line 649: FWIW I really prefer frequency polygons to density plots.

If you do keep density plots, I think you need to say something like "a density plot is a type of smoothed histogram" and provide some justification as to why you'd use it.

data-visualize.qmd
```diff
@@ -651,36 +646,32 @@ ggplot(penguins, aes(x = species, y = body_mass_g)) +
...
 
 ggplot(penguins, aes(x = body_mass_g, color = species)) +
-  geom_freqpoly(binwidth = 200, linewidth = 0.75)
+  geom_density(linewidth = 0.75)
 ```
 
 We've also customized the thickness of the lines using the `linewidth` argument in order to make them stand out a bit more against the background.
 
-We can also use overlaid density plots, with `species` mapped to both `color` and `fill` aesthetics and using the `alpha` aesthetic to add transparency to the filled density curves.
+We can also map `species` to both `color` and `fill` aesthetics and using the `alpha` aesthetic to add transparency to the filled density curves.
```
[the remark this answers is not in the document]
hadley: And in the next sentence, I think you need to say "The alpha aesthetic" instead of "This aesthetic"

data-visualize.qmd
```diff
@@ -778,7 +774,7 @@ You will learn about many other geoms for visualizing distributions of variables
 
 ### Exercises
 
-1.  Which variables in `mpg` are categorical?
+1.  Which variables in `mpg` are categorical in the context of the data?
```
hadley, line 777: In what other context would you consider them?

intro.qmd
```diff
@@ -35,15 +35,14 @@ knitr::include_graphics("diagrams/data-science/base.png", dpi = 270)
 
 First, you must **import** your data into R.
 This typically means that you take data stored in a file, database, or web application programming interface (API) and load it into a data frame in R.
-If you can't get your data into R, you can't do data science on it!
+If you can't get your data into R, you can't do data science on it with R!
```
hadley, line 38:
```suggestion
If you can't load your data, you can't do data science on it!
```

intro.qmd
```diff
@@ -64,20 +63,20 @@ That means a model cannot fundamentally surprise you.
...
 
 ## How this book is organized
 
 The previous description of the tools of data science is organized roughly according to the order in which you use them in an analysis (although, of course, you'll iterate through them multiple times).
-In our experience, however, learning data ingesting and tidying first is sub-optimal because 80% of the time, it's routine and boring, and the other 20% of the time, it's weird and frustrating.
-That's a bad place to start learning a new subject!
-Instead, we'll start with visualization and transformation of data that's already been imported and tidied.
+In our experience, however, learning data ingesting and tidying first is sub-optimal because in many cases it's a fairly easy task, and we don't want to get bogged down in the complicated cases for now.
+
+Instead, we'll start with imported and tidied data, so we can go straight to making interesting insights through visualization and transformation.
```
hadley, line 79:
```suggestion
Instead, we'll start with pre-loaded tidy data, so we can go straight to making interesting insights through visualization and transformation.
```

hadley (approved):
You did get a lot more feedback in these chapters!

data-visualize.qmd
```diff
@@ -732,7 +728,7 @@ ggplot(penguins, aes(x = flipper_length_mm, y = body_mass_g)) +
 
 ### Three or more variables
 
-One way to add additional variables to a plot is by mapping them to an aesthetic.
+As we saw in our Ultimate Goal, we can incorporate more variables into a plot by mapping them to additional aesthetics.
```
hadley, line 731: You used "ultimate goal" in other places

data-visualize.qmd
mine-cetinkaya-rundel: Do you prefer quantitative? I don't mind quantitative, but I don't like qualitative > categorical, and I don't like numeric + categorical or quantitative + categorical together, so I stuck with numerical + categorical pair. Do you have a specific preference for pair of terms?

data-visualize.qmd
mine-cetinkaya-rundel: It is referring to that plot, but maybe not necessary to do so.

data-visualize.qmd
hadley: I can just live with numerical.

data-visualize.qmd
mine-cetinkaya-rundel: I agree, I don't think this is an improvement. It was an attempt to distinguish between R thinking a 0/1 variable could be numerical when it's actually categorical. But we haven't introduced data types yet here so I'm going to revert this, I think it complicates instead of clarifying.

PR #1331, opened 2023-03-01
Updates for new relationship argument

### LINKED ISSUE #1299: Update joins for new relationship argument
https://github.com/tidyverse/dplyr/pull/6753/files

hadley:
Since we need to reduce the number of pages, and most folks will no longer need be exposed to the `multiple` argument, I reduced the content significantly.

@davisvaughan can you take a look too please?

Fixes #1299

### DIFF
--- a/joins.qmd
+++ b/joins.qmd
@@ -412,8 +412,7 @@ flights2 |>
 ## How do joins work?
 
 Now that you've used joins a few times it's time to learn more about how they work, focusing on how each row in `x` matches rows in `y`.
-We'll begin by using @fig-join-setup to introduce a visual representation of the two simple tibbles defined below.
-In these examples we'll use a single key called `key` and a single value column (`val_x` and `val_y`), but the ideas all generalize to multiple keys and multiple values.
+We'll begin by introducing a visual representation of joins, using the simple tibbles defined below and shown in @fig-join-setup. In these examples we'll use a single key called `key` and a single value column (`val_x` and `val_y`), but the ideas all generalize to multiple keys and multiple values.
 
 ```{r}
 x <- tribble(
@@ -446,7 +445,8 @@ y <- tribble(
 knitr::include_graphics("diagrams/join/setup.png", dpi = 270)
 ```
 
-@fig-join-setup2 shows all potential matches between `x` and `y` as the intersection between lines drawn from each row of `x` and each row of `y`.
+@fig-join-setup2 introduces the foundation for our visual representation.
+It shows all potential matches between `x` and `y` as the intersection between lines drawn from each row of `x` and each row of `y`.
 The rows and columns in the output are primarily determined by `x`, so the `x` table is horizontal and lines up with the output.
 
 ```{r}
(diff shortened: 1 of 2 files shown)


joins.qmd
```diff
@@ -465,8 +465,9 @@ The rows and columns in the output are primarily determined by `x`, so the `x` t
 knitr::include_graphics("diagrams/join/setup2.png", dpi = 270)
 ```
 
-In an actual join, matches will be indicated with dots, as in @fig-join-inner.
-The number of dots equals the number of matches, which in turn equals the number of rows in the output, a new data frame that contains the key, the x values, and the y values.
+To describe a specific type of join, we indicate matches with dots.
+The matches determines the rows in output, a new data frame that contains the key, the x values, and the y values.
```
DavisVaughan, line 469:
```suggestion
The matches determine the rows in the output, a new data frame that contains the key, the x values, and the y values.
```

joins.qmd
```diff
@@ -465,8 +465,9 @@ The rows and columns in the output are primarily determined by `x`, so the `x` t
...
+For example, @fig-join-inner shows an inner join, where two rows match if and only if the keys are equal.
```
DavisVaughan, line 470:
```suggestion
For example, @fig-join-inner shows an inner join, where rows are retained if and only if the keys are equal.
```

joins.qmd
```diff
@@ -484,7 +485,7 @@ The number of dots equals the number of matches, which in turn equals the number
 knitr::include_graphics("diagrams/join/inner.png", dpi = 270)
 ```
 
-An **outer join** keeps observations that appear in at least one of the data frames.
+We can apply the same principles to explain the **outer joins**, which keep observations that appear in at least one of the data frames.
```
DavisVaughan, line 488: I don't know if this is valuable, but I've enjoyed thinking of outer joins as performing _exactly_ the same matching procedure as an inner join - the only difference is in how they handle _unmatched_ rows on either side of the join.

i.e. the base of an outer join is an inner join, patched up with some of the unmatched rows

DavisVaughan: I guess this is different from how you talk about it though, because you say that there is a catch-all virtual row that matches against "anything" that isn't already handled. So that takes the place of talking about handling _unmatched_ rows, because in your explanation there aren't any anymore (because the catch all handled them already)

joins.qmd
```diff
@@ -606,78 +607,26 @@ There are three possible outcomes for a row in `x`:
...
-df2 <- tibble(x = c(1, 1))
-df3 <- tibble(x = 3)
-
 df1 |> 
-  inner_join(df2, join_by(x), unmatched = "error", multiple = "error")
-df1 |> 
-  inner_join(df3, join_by(x), unmatched = "error", multiple = "error")
-```
-
-Note that `unmatched = "error"` is not useful with `left_join()` because, as described above, every row in `x` has a fallback match to a virtual row in `y`.
```
DavisVaughan, line 650: Not exactly true anyways, for `left_join()` it will error if a row in `y` was about to be dropped

It isn't useful for `full_join()` (all rows are always kept), so it isn't an argument to that join function.

DavisVaughan (approved):
I think this is a very good improvement and targets the most important problem, many-to-many

joins.qmd
mine-cetinkaya-rundel: I would keep the description as is but remind by appending ", unlike an inner join that keeps observations that appear in both tables."

mine-cetinkaya-rundel (approved):

PR #1335, opened 2023-03-02
More TR edits

mine-cetinkaya-rundel:


data-transform.qmd
```diff
@@ -63,16 +62,16 @@ These are important because the operations you can perform on a column depend so
...
 
 1.  The first argument is always a data frame.
 
-2.  The subsequent arguments describe what to do with the data frame, using the variable names (without quotes).
+2.  The subsequent arguments typically describe which columns to operate on, using the variable names (without quotes).
 
 3.  The result is always a new data frame.
 
-Since each verb is quite simple, solving complex problems will usually require combining multiple verbs, and we'll do so with the pipe, `|>`.
+Since the functionality of each verb is limited, solving complex problems will usually require combining multiple verbs, and we'll do so with the pipe, `|>`.
```
hadley, line 74: What about "Because each verb does one thing well" ?

data-transform.qmd
```diff
@@ -166,29 +165,32 @@ flights |>
...
+We get the earliest years first, then within a year the earliest months, etc.
 
 ```{r}
 flights |> 
   arrange(year, month, day, dep_time)
 ```
 
-You can use `desc()` to re-order by a column in descending order.
-For example, this code shows the most delayed flights:
+You can use `desc()` to re-order the data frame based a column, in descending order.
```
hadley, line 184:
```suggestion
You can use `desc()` to re-order the data frame based a column in descending order.
```

But while the previous text was technically incorrect, I think it gets the message across more clearly.

data-transform.qmd
```diff
@@ -166,29 +165,32 @@ flights |>
...
+For example, this code shows the most delayed flights on top:
```
hadley, line 185:
```suggestion
For example, this code shows the most delayed flights first:
```

data-transform.qmd
```diff
@@ -548,13 +561,14 @@ In dplyr, this is operation is performed by `summarize()`[^data-transform-3], as
...
+    avg_delay = mean(dep_delay)
   )
 ```
 
 Uhoh!
-Something has gone wrong and all of our results are `NA` (pronounced "N-A"), R's symbol for missing value.
-We'll come back to discuss missing values in detail in @sec-missing-values, but for now we'll remove them by using `na.rm = TRUE`:
+Something has gone wrong and all of our results are `NA`s (pronounced "N-A"), R's symbol for missing value.
+This happened because some of the observed flights had missing data in the delay column, and so when we calculated the mean including those values, we got an `NA` result.
+We'll come back to discuss missing values in detail in @sec-missing-values, but for now we'll tell the `mean()` function to ignore all missing values by adding the setting the argument `na.rm` to `TRUE`:
```
hadley, line 571:
```suggestion
We'll come back to discuss missing values in detail in @sec-missing-values, but for now we'll tell the `mean()` function to ignore all missing values by setting the argument `na.rm` to `TRUE`:
```

data-transform.qmd
```diff
@@ -637,28 +650,34 @@ Alternatively, change the default behavior by setting a different value, e.g. `"
...
+You might also want to remove grouping from a data frame without using `summarize()`.
 You can do this with `ungroup()`.
 
 ```{r}
 daily |> 
-  ungroup() |>
+  ungroup()
+```
+
+Now let's summarize see whathappens when you summarize an ungrouped data frame.
```
hadley, line 661:
```suggestion
Now let's see what happens when you summarize an ungrouped data frame.
```

data-transform.qmd
```diff
@@ -678,8 +697,7 @@ As you can see, when you summarize an ungrouped data frame, you get a single row
     )
     ```
 
-    a.  What does the following code do?
-        Run it, analyze the result, and describe what `group_by()` does.
+    a.  Write down what you think the output will look like, then check if you were correct, and describe what `group_by()` does.
```
hadley, line 700: Make this an unordered list to match the exercises you changed above?

data-visualize.qmd
```diff
@@ -83,7 +83,7 @@ A data frame is a rectangular collection of variables (in the columns) and obser
 
 Type the name of the data frame in the console and R will print a preview of its contents.
 Note that it says `tibble` on top of this preview.
-In the tidyverse, we use special data frames called tibbles that you will learn more about soon.
+In the tidyverse, we use special data frames called **tibbles** that you will learn more about soon.
```
hadley, line 86: Do we talk more about tibbles?

hadley (approved):

data-transform.qmd
mine-cetinkaya-rundel: > You can use `desc()` to re-order the data frame by a column in descending order.

This seems like a good middle ground.

data-visualize.qmd
mine-cetinkaya-rundel: Yup, we do at https://r4ds.hadley.nz/data-transform.html#nycflights13.

data-transform.qmd
mine-cetinkaya-rundel: These are parts (a), (b), (c), etc., should be completed consecutively, vs. the other one was filtering for all of those conditions at once (or at least I thought so). If not, I should bring the other one back.

PR #1338, opened 2023-03-03
Mynor typo

davidrsch:
Changing 2 with 10 so in the line.
"Then, type the name of the second argument, `to`, and set it equal to `10`" so it is relate to example:
```
seq(from = 1, to = 10)
```


mine-cetinkaya-rundel:
Thank you!

PR #1339, opened 2023-03-04
Minors

oliviercailloux:
Started reading this great book, thanks for it. Here are a few minor things I spotted in the first four chapters while reading. And a few TODOs in the code with suggestions, feel free to disregard them if they are misguided.


intro.qmd
```diff
@@ -86,9 +86,8 @@ Although it can be tempting to skip the exercises, there's no better way to lear
 
 ## What you won't learn
 
-There are several important topics that this book doesn't cover.
-We believe it's important to stay ruthlessly focused on the essentials so you can get up and running as quickly as possible.
-That means this book can't cover every important topic.
+There are several important topics that this book doesn't cover:
+we believe it's important to stay ruthlessly focused on the essentials so you can get up and running as quickly as possible.
```
mine-cetinkaya-rundel, line 90:
```suggestion
We believe it's important to stay ruthlessly focused on the essentials so you can get up and running as quickly as possible.
That means this book can't cover every important topic.
```

intro.qmd
```diff
@@ -86,7 +86,6 @@ Although it can be tempting to skip the exercises, there's no better way to lear
 
 ## What you won't learn
 
-There are several important topics that this book doesn't cover.
 We believe it's important to stay ruthlessly focused on the essentials so you can get up and running as quickly as possible.
```
mine-cetinkaya-rundel, line 89:
```suggestion
There are several important topics that this book doesn't cover.
We believe it's important to stay ruthlessly focused on the essentials so you can get up and running as quickly as possible.
```

mine-cetinkaya-rundel:
Thank you!

PR #1348, opened 2023-03-08
Visualize part review edits

mine-cetinkaya-rundel:
And other parts moving around text has affected


EDA.qmd
```diff
@@ -290,13 +266,6 @@ If you've encountered unusual values in your dataset, and simply want to move on
       mutate(y = if_else(y < 3 | y > 20, NA, y))
     ```
 
-`if_else()` has three arguments.
```
hadley, line 293: IMO this is the biggest benefit of the TRs — spotting when we're moved stuff so we accidentally use it before we teach it.

EDA.qmd
```diff
@@ -278,7 +254,7 @@ If you've encountered unusual values in your dataset, and simply want to move on
       filter(between(y, 3, 20))
     ```
 
-    We don't recommend this option because just because one measurement is invalid, doesn't mean all the measurements are.
+    We don't recommend this option because, for a given observation, just because the value for one variable is invalid doesn't mean the values for all variables are.
```
hadley, line 257:
```suggestion
    We don't recommend this option because one invalid value doesn't imply that all the other values for that observation are also invalid.
```

EDA.qmd
```diff
@@ -379,17 +352,10 @@ ggplot(diamonds, aes(x = price)) +
...
-```{r}
-#| fig-alt: >
-#|   Bar chart of cuts of diamonds showing large variability between the 
-#|   frequencies of various cuts. Fair diamonds have the lowest frequency, 
-#|   then Good, then Very Good, then Premium, and then Ideal.
-
-ggplot(diamonds, aes(x = cut)) + 
-  geom_bar()
-```
+The default appearance of `geom_freqpoly()` is not that useful for that sort of comparison because the height is given by the count and the overall counts of `cut` in differ so much, making it hard to see the differences in the shapes of their distributions.
```
hadley, line 358:
```suggestion
The default appearance of `geom_freqpoly()` is not that useful here because the height, determined by the overall count, differs so much across `cut`s, making it hard to see the differences in the shapes of their distributions.
```

EDA.qmd
```diff
@@ -575,7 +540,7 @@ ggplot(smaller, aes(x = carat, y = price)) +
 
 (In this section we'll use the `smaller` dataset to stay focused on the bulk of the diamonds that are smaller than 3 carats)
 
-Scatterplots become less useful as the size of your dataset grows, because points begin to overplot, and pile up into areas of uniform black (as above).
+Scatterplots become less useful as the size of your dataset grows, because points begin to overplot, and pile up into areas of uniform black, making it hard to see the trend.
```
hadley, line 543:
```suggestion
Scatterplots become less useful as the size of your dataset grows, because points begin to overplot, and pile up into areas of uniform black, making it hard to see where distribution in space.
```

That's maybe not quite right either, but it's definitely the 2d distribution that it affects, not just the ability to judge trend.

communication.qmd
```diff
@@ -85,35 +62,21 @@ ggplot(mpg, aes(x = displ, y = hwy)) +
...
+    color = "Car type",
     title = "Fuel efficiency generally decreases with engine size",
     subtitle = "Two seaters (sports cars) are an exception because of their light weight",
     caption = "Data from fueleconomy.gov"
   )
 ```
 
+The purpose of a plot title is to summarize the main finding.
+Avoid titles that just describe what the plot is, e.g. "A scatterplot of engine displacement vs. fuel economy".
+If you need to add more text, there are two other useful labels: `subtitle` adds additional detail in a smaller font beneath the title and `caption` adds text at the bottom right of the plot, often used to describe the source of the data.
```
hadley, line 76:
```suggestion
Avoid titles that just describe what the plot is, e.g. "A scatterplot of engine displacement vs. fuel economy".

If you need to add more text, there are two other useful labels: `subtitle` adds additional detail in a smaller font beneath the title and `caption` adds text at the bottom right of the plot, often used to describe the source of the data.
```

communication.qmd
```diff
@@ -280,69 +220,47 @@ ggplot(mpg, aes(x = displ, y = hwy)) +
...
+    We often make them thick (`linewidth = 2`) and white (`color = white`), and draw them underneath the primary data layer.
+    That makes them easy to see, without drawing attention away from the data.
 
-label_info <- mpg |>
-  summarize(
-    displ = max(displ),
-    hwy = max(hwy),
-    label = "Increasing engine size is \nrelated to decreasing fuel economy."
-  )
+-   Use `geom_rect()` to draw a rectangle around points of interest.
```
hadley, line 230: Maybe mention https://ggforce.data-imaginist.com/reference/geom_mark_hull.html ?

communication.qmd
```diff
@@ -753,7 +659,7 @@ ggplot(df, aes(x, y)) +
   labs(title = "Viridis, binned")
 ```
 
-Note that all color scales come in two variety: `scale_color_x()` and `scale_fill_x()` for the `color` and `fill` aesthetics respectively (the color scales are available in both UK and US spellings).
+Note that all color scales come in two variety: `scale_color_*()` and `scale_fill_*()` for the `color` and `fill` aesthetics respectively (the color scales are available in both UK and US spellings).
```
hadley, line 662:
```suggestion
Note that all color scales come in two varieties: `scale_color_*()` and `scale_fill_*()` for the `color` and `fill` aesthetics respectively (the color scales are available in both UK and US spellings).
```

data-import.qmd
```diff
@@ -151,6 +151,14 @@ students <- students |>
 students
 ```
 
+A new function here is `if_else()`, which has three arguments.
+The first argument `test` should be a logical vector.
+The result will contain the value of the second argument, `yes`, when `test` is `TRUE`, and the value of the third argument, `no`, when it is `FALSE`.
+Here we're saying if `age` is the character string `"five"`, make it `"5"`, and if not leave it as `age`.
+Alternatively to `if_else()`, use `case_when()`.
+`case_when()` is particularly useful inside mutate when you want to create a new variable that relies on a complex combination of existing variables or would otherwise require multiple `if_else()` statements nested inside one another.
```
hadley, line 159:
```suggestion
```

data-import.qmd
```diff
@@ -151,6 +151,14 @@ students <- students |>
...
+You will learn more about logical vectors in @sec-logicals.
```
hadley, line 160:
```suggestion
You will learn more about `if_else()` and logical vectors in @sec-logicals.
```

layers.qmd
```diff
@@ -176,7 +172,7 @@ You'll need to pick a value that makes sense for that aesthetic:
 #|   point-up blue, 25 - filled triangle point down blue.
 
 shapes <- tibble(
```
hadley, line 174: That ordered was carefully picked to put similar shapes next to each other. I'd prefer to leave it as is, and call out the non-ascending order in the legend.

layers.qmd
```diff
@@ -265,7 +261,7 @@ ggplot(mpg, aes(x = displ, y = hwy)) +
   geom_smooth()
 ```
 
-Every geom function in ggplot2 takes a `mapping` argument.
+Every geom function in ggplot2 takes a `mapping` argument, which it inherits from `ggplot()`.
```
hadley, line 264: I don't really get what that means. Inherits in what sense?

layers.qmd
```diff
@@ -338,7 +333,7 @@ ggplot(mpg, aes(x = displ, y = hwy)) +
   geom_smooth(aes(group = drv))
     
 ggplot(mpg, aes(x = displ, y = hwy)) +
-  geom_smooth(aes(color = drv), show.legend = FALSE)
+  geom_smooth(aes(color = drv))
```
hadley, line 336: I suspect that legend was deliberately omitted to give more space for the data

layers.qmd
```diff
@@ -751,8 +738,9 @@ Each stat is a function, so you can get help in the usual way, e.g. `?stat_bin`.
     How is it different from `geom_bar()`?
 
 3.  Most geoms and stats come in pairs that are almost always used in concert.
-    Read through the documentation and make a list of all the pairs.
+    Make a list of all the pair.
```
hadley, line 741:
```suggestion
    Make a list of all the pairs.
```

layers.qmd
```diff
@@ -751,8 +738,9 @@ Each stat is a function, so you can get help in the usual way, e.g. `?stat_bin`.
...
     What do they have in common?
+    (Hint: Read through the documentation to find the pairs.)
```
hadley, line 743:
```suggestion
    (Hint: Read through the documentation.)
```

hadley (approved):

layers.qmd
mine-cetinkaya-rundel: Referring to global mapping vs. local mapping, I'll clarify

EDA.qmd
mine-cetinkaya-rundel: Agreed, those have been super helpful!

EDA.qmd
mine-cetinkaya-rundel: Agreed, replacing with 

> making it hard to judge differences in the density of the data across the 2-dimonsional space as well as making it hard to spot the trend.

layers.qmd
mine-cetinkaya-rundel: I figured as much. I think both approaches make sense, but I reverted to the original.

PR #1351, opened 2023-03-09
There's something about `.by`

### LINKED ISSUE #1242: Say something about `.by`
and link to https://dplyr.tidyverse.org/dev/reference/dplyr_by.html

hadley:
Fixes #1242

### DIFF
--- a/data-transform.qmd
+++ b/data-transform.qmd
@@ -699,6 +699,38 @@ daily |>
 
 You get a single row back because dplyr treats all the rows in an ungrouped data frame as belonging to one group.
 
+### `.by`
+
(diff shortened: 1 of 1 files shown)


mine-cetinkaya-rundel (approved):
This is great, the only comment I have is whether we should add a sentence around this being an "alternative" syntax and group_by() approach not going away. I feel like that would be useful.

If adopting the word "alternative" the beginning of the section can say "dplyr introduces an alternative syntax" instead of "new" syntax.

hadley:
I felt like alternative was more likely to incite worry, but I now emphasise that `group_by()` and `ungroup()` aren't going away.

PR #1354, opened 2023-03-09
Adjust multi-column plots

hadley:


mine-cetinkaya-rundel:
@hadley I can't tag you as the reviewer since you started the PR, but take a look. In some cases, I changed the dataset being used so the labels could fit without having to increase `fig-width` too much, making the text too small.

hadley:
Thanks! I was stuck on that problem.

PR #1357, opened 2023-03-10
Replace with realistic data for "how does `pivot_*()` work"

mine-cetinkaya-rundel:
This PR will require updating the following diagrams:

- tidy-data/variables
- tidy-data/column-names
- tidy-data/cell-values

I've already updated the alt text for them in the PR.

Also, if we need more space, we could cut 

> You might also wonder what happens if there are multiple rows in the input that correspond to one cell in the output.

onwards in "How does `pivot_wider() work?"


hadley (approved):
Diagrams updated

mine-cetinkaya-rundel:
@hadley I wrote up the text replacing `var` for `id`. Is it a huge pain to update those in the diagrams?

hadley:
Oh I just missed that. I'll update.

PR #1363, opened 2023-03-12
Update acknowledgements

hadley:


hadley:
I've merge so I can upload another version to O'Reilly, but we can still change if need.

PR #1364, opened 2023-03-13
Add `distance` to line 301

thefoxboxman:
.keep=Used argument should keep the `distance` variable because it is used in the mutate() step.


thefoxboxman:
Sorry I read it wrong. My bad. Cheers

PR #1367, opened 2023-03-15
Replaced group_by() with .by

KittJonathan:


mine-cetinkaya-rundel:
We have decided to mention `.by` at https://r4ds.hadley.nz/data-transform.html#by but not add examples based on it elsewhere in the book, at least not in this edition. So I'll close this PR but thanks for the suggestion!

PR #1372, opened 2023-03-16
Fix typo in intro.qmd

kevinxperese:
Plus add a little more clarification/rationale for installing the additional packages.


hadley:
Thanks!

PR #1383, opened 2023-03-22
Wrong dataset fix

alberto-agudo:
This exercise asks to create a decimal time out of clock time from the `sched_dep_time` variable. However, it is calling this on the `weather` dataset, where it should be the `flights` one. Otherwise, the user will get an error.


hadley:
Thanks!

PR #1384, opened 2023-03-22
Including msising word

alberto-agudo:
I am not completely sure about this one, but I guess the real intention was saying that when you use `across` you are **not** calling the new function yourself.


hadley:
Thanks!

@mine-cetinkaya-rundel this is a fix we'll want to port to the published version. Can you remind me how we do that?

PR #1391, opened 2023-03-27
Fix/workflow-basics

mitsuoxv:
justify omitting the name of the second argument

### DIFF
--- a/workflow-basics.qmd
+++ b/workflow-basics.qmd
@@ -180,7 +180,7 @@ Finally, hit return.
 seq(from = 1, to = 10)
 ```
 
-We often omit the names of the first arguments in function calls, so we can rewrite this as follows:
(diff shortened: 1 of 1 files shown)


workflow-basics.qmd
```diff
@@ -180,7 +180,7 @@ Finally, hit return.
 seq(from = 1, to = 10)
 ```
 
-We often omit the names of the first arguments in function calls, so we can rewrite this as follows:
+We often omit the names of the first several arguments whose order is unambiguous in function calls, so we can rewrite this as follows:
```
mine-cetinkaya-rundel, line 183:
```suggestion
We often omit the names of the first several arguments in function calls, so we can rewrite this as follows:
```

mine-cetinkaya-rundel:
I adjusted the suggestion a bit -- the order may not be unambiguous to everyone -- but the addition of "several" helps. Thank you!

PR #1398, opened 2023-03-30
Fix/data-transform

mitsuoxv:
fix wrong references, inconsistency between sentence and code, and typos

### DIFF
--- a/data-transform.qmd
+++ b/data-transform.qmd
@@ -251,7 +251,7 @@ flights |>
 
 ## Columns
 
-There are four important verbs that affect the columns without changing the rows: `mutate()` creates new columns that are derived from the existing columns; `select()` changes which columns are present; `rename()` changes the names of the columns; and `relocate()` changes the positions of the columns.
+There are four important verbs that affect the columns without changing the rows: `mutate()` creates new columns that are derived from the existing columns, `select()` changes which columns are present, `rename()` changes the names of the columns, and `relocate()` changes the positions of the columns.
 
 ### `mutate()` {#sec-mutate}
 
--- a/data-transform.qmd
+++ b/data-transform.qmd
@@ -58,7 +58,7 @@ glimpse(flights)
 ```
 
 In both views, the variables names are followed by abbreviations that tell you the type of each variable: `<int>` is short for integer, `<dbl>` is short for double (aka real numbers), `<chr>` for character (aka strings), and `<dttm>` for date-time.
-These are important because the operations you can perform on a column depend so much on its "type", and these types are used to organize the chapters in the next section of the book.
+These are important because the operations you can perform on a column depend so much on its "type", and these types are used to organize the chapters in the Transform part of the book.
 
 ### dplyr basics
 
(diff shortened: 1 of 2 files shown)


data-transform.qmd
```diff
@@ -58,7 +58,7 @@ glimpse(flights)
 ```
 
 In both views, the variables names are followed by abbreviations that tell you the type of each variable: `<int>` is short for integer, `<dbl>` is short for double (aka real numbers), `<chr>` for character (aka strings), and `<dttm>` for date-time.
-These are important because the operations you can perform on a column depend so much on its "type", and these types are used to organize the chapters in the next section of the book.
+These are important because the operations you can perform on a column depend so much on its "type", and these types are used to organize the chapters in the Transform part of the book.
```
mine-cetinkaya-rundel, line 61:
```suggestion
These are important because the operations you can perform on a column depend so much on its "type".
```

data-transform.qmd
```diff
@@ -166,7 +166,7 @@ flights |>
 ```
 
 This "works", in the sense that it doesn't throw an error, but it doesn't do what you want because `|` first checks the condition `month == 1` and then checks the condition `2`, which is not a sensible condition to check.
-We'll learn more about what's happening here and why in @sec-boolean-operations.
+We'll learn more about what's happening here and why in @sec-order-of-operations-logicals.
```
mine-cetinkaya-rundel, line 169:
```suggestion
We'll learn more about what's happening here and why in @sec-boolean-operations.
```

data-transform.qmd
```diff
@@ -251,7 +251,7 @@ flights |>
 
 ## Columns
 
-There are four important verbs that affect the columns without changing the rows: `mutate()` creates new columns that are derived from the existing columns, `select()` changes which columns are present; `rename()` changes the names of the columns; and `relocate()` changes the positions of the columns.
+There are four important verbs that affect the columns without changing the rows: `mutate()` creates new columns that are derived from the existing columns; `select()` changes which columns are present; `rename()` changes the names of the columns; and `relocate()` changes the positions of the columns.
```
mine-cetinkaya-rundel, line 254:
```suggestion
There are four important verbs that affect the columns without changing the rows: `mutate()` creates new columns that are derived from the existing columns, `select()` changes which columns are present, `rename()` changes the names of the columns, and `relocate()` changes the positions of the columns.
```

logicals.qmd
```diff
@@ -247,7 +247,7 @@ A missing value in a logical vector means that the value could either be `TRUE`
 `TRUE | TRUE` and `FALSE | TRUE` are both `TRUE`, so `NA | TRUE` must also be `TRUE`.
 Similar reasoning applies with `NA & FALSE`.
 
-### Order of operations
+### Order of operations {#sec-order-of-operations-logicals}
```
mine-cetinkaya-rundel, line 250:
```suggestion
### Order of operations
```

mine-cetinkaya-rundel:
Thank you!

PR #1408, opened 2023-04-02
Suggestion/introduction

mitsuoxv:
use real examples, and show we can access an object in the package without loading it


mine-cetinkaya-rundel:
Thank you for the suggestion! I think the examples work as is so I'll close this without merging, but I appreciate the suggestion.

PR #1410, opened 2023-04-02
Update strings.qmd

alansuidaihe:
MInor


mine-cetinkaya-rundel:
Thank you! However, throughout the book, we don't use `code font` for package names, per https://style.tidyverse.org/documentation.html?q=package#r-code, so I'll close this without merging.

PR #1418, opened 2023-04-10
Fix name in pivotting column

### LINKED ISSUE #1404: Duplicated column names in Figure 6.7 in data-tidy
I find column name "id" appears twice in the right data frame of Figure 6.7 in data-tidy.qmd. Better to change `names_to = c(".value", "id")` to something like `names_to = c(".value", "number")`.

hadley:
Fixes #1404


mine-cetinkaya-rundel:
Thanks, and updated in the sheet!

PR #1430, opened 2023-04-17
Fix typo in missing-values

petzi53:


mine-cetinkaya-rundel:
I believe "it" is correct here, it's one dataset.

missing-values.qmd
```diff
@@ -176,7 +176,7 @@ In that case, you can do manually what `complete()` does for you: create a data
 ### Joins
 
 This brings us to another important way of revealing implicitly missing observations: joins.
-You'll learn more about joins in @sec-joins, but we wanted to quickly mention them to you here since you can often only know that values are missing from one dataset when you compare it another.
+You'll learn more about joins in @sec-joins, but we wanted to quickly mention them to you here since you can often only know that values are missing from one dataset when you compare them to another.
```
mine-cetinkaya-rundel, line 179:
```suggestion
You'll learn more about joins in @sec-joins, but we wanted to quickly mention them to you here since you can often only know that values are missing from one dataset when you compare it to another.
```

mine-cetinkaya-rundel:
Thank you!

PR #1439, opened 2023-04-19
Fix/communication probably typos

mitsuoxv:
- no need for "\n" before str_wrap()
- a typo, probably forgotten to delete
- "col" can work, but may be confusing


mine-cetinkaya-rundel:
Thank you!

PR #1440, opened 2023-04-20
Fix/logicals mainly typos

mitsuoxv:
make codes consistent with sentences, and typos


mine-cetinkaya-rundel:
Thank you for these! I've implemented things a bit differently -- matching text to code instead of the other way around -- as that was easier to deal with at this stage of the print book editing.

PR #1443, opened 2023-04-24
Fix curl::multi_download typo in arrow.qmd

nstjhp:
Function has an underscore in the name (as used in following code box)


hadley:
Thanks!

PR #1449, opened 2023-04-28
Eliminate infeasible step in exercise in Ch 20

ganong123:
The first sentence in exercise 20.3.4 #5 seems to be infeasible with the tools that have been taught in the textbook.

@rntucker proposed this solution which is correct, but relies on tools not in the textbook. 
```
plane_carriers <- nycflights13::planes |>
  left_join(flights |> select(tailnum, carrier), by = "tailnum") |>
  distinct(carrier, tailnum) |>
  group_by(tailnum) |>
  summarise(all_carriers = toString(carrier), .groups = "drop") 

plane_carriers %>% count(all_carriers)
```

I suggest reverting to the language from [1e](https://r4ds.had.co.nz/relational-data.html) in which case this much simpler solution works
```
flights |>
  distinct(carrier, tailnum) |>
  count(tailnum) |>
  filter(n > 1)
```

### DIFF
--- a/joins.qmd
+++ b/joins.qmd
@@ -367,8 +367,7 @@ flights2 |>
 4.  What do the tail numbers that don't have a matching record in `planes` have in common?
     (Hint: one variable explains \~90% of the problems.)
 
-5.  Add a column to `planes` that lists every `carrier` that has flown that plane.
-    You might expect that there's an implicit relationship between plane and airline, because each plane is flown by a single airline.
+5.  You might expect that there's an implicit relationship between plane and airline, because each plane is flown by a single airline.
     Confirm or reject this hypothesis using the tools you've learned in previous chapters.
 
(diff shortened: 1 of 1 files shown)


hadley:
Doesn't this work?

```R
library(tidyverse)
library(nycflights13)

planes |>
  left_join(flights, by = "tailnum") |>
  distinct(tailnum, carrier) |> 
  group_by(tailnum) |>
  summarise(
    all_carriers = str_flatten(carrier, collapse = ","),
    .groups = "drop"
  ) 
```

ganong123:
@Hadley yes, I think your solution works! I have never used `str_flatten()` before so I didn't think of it. 

I do think the question would be better with a hint since the use of `str_flatten()` doesn't seem obvious at all, at least to me.

PR #1469, opened 2023-05-12
Suggest/joins

mitsuoxv:
- better to make code compatible with the figure
- better to remove the last row: 1 Max NA
- make sure at least one employee born before Jan 10, and make output reproducible (the current output shows no employee born before Jan 10 as of today: May 12, 2023)


mine-cetinkaya-rundel:
Thank you!

PR #1472, opened 2023-05-15
Fix/Spreadsheets probably typos

mitsuoxv:
probably typos

### DIFF
--- a/spreadsheets.qmd
+++ b/spreadsheets.qmd
@@ -170,7 +170,7 @@ However there are a few things we might want to address in this dataset:
     ```
 
 It took us multiple steps and trial-and-error to load the data in exactly the format we want, and this is not unexpected.
-Data science is an iterative process, and the process of iteration can be even more tedious when reading data in from spreadsheets compared to other plain text rectangular data files because humans tend to input data into spreadsheets and use them not just for data storage but also for sharing and communication.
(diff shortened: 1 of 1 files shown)


spreadsheets.qmd
```diff
@@ -170,7 +170,7 @@ However there are a few things we might want to address in this dataset:
     ```
 
 It took us multiple steps and trial-and-error to load the data in exactly the format we want, and this is not unexpected.
-Data science is an iterative process, and the process of iteration can be even more tedious when reading data in from spreadsheets compared to other plain text, rectangular data files because humans tend to input data into spreadsheets and use them not just for data storage but also for sharing and communication.
+Data science is an iterative process, and the process of iteration can be even more tedious when reading data in from spreadsheets compared to other plain text rectangular data files because humans tend to input data into spreadsheets and use them not just for data storage but also for sharing and communication.
```
mine-cetinkaya-rundel, line 173:
```suggestion
Data science is an iterative process, and the process of iteration can be even more tedious when reading data in from spreadsheets compared to other plain text, rectangular data files because humans tend to input data into spreadsheets and use them not just for data storage but also for sharing and communication.
```

mine-cetinkaya-rundel, line 173:
```suggestion
Data science is an iterative process, and the process of iteration can be even more tedious when reading data in from spreadsheets compared to other plain text, rectangular data files because humans tend to input data into spreadsheets and use them not just for data storage but also for sharing and communication.
```

mine-cetinkaya-rundel:
Thank you!

PR #1475, opened 2023-05-16
Update data-transform.qmd, fixed sentence

andyhuynh92:
The original sentence read "We didn't focus on this syntax in this chapter because it was very new when wrote the book. ", and I added a missing subject, so it now reads "We didn't focus on this syntax in this chapter because it was very new when we wrote the book."


mine-cetinkaya-rundel:
Thank you!

PR #1477, opened 2023-05-17
Suggest/Strings, 15.4.3 Diagnosing widening problems, double "x"

mitsuoxv:
Input column is "x" and output columns are c("x", "y", "z), so "x" is duplicated.
I would like to suggest to replace names = c("x", "y", "z) with c("a", "b", "c").

### DIFF
--- a/strings.qmd
+++ b/strings.qmd
@@ -356,7 +356,7 @@ df |>
   separate_wider_delim(
     x,
     delim = "-",
-    names = c("x", "y", "z")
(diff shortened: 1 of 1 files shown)


mine-cetinkaya-rundel:
Thank you for the suggestion. I think the existing example is okay, and I'm worried making this change now will create a big divergence from the print version in a way that doesn't necessarily add a huge amount clarity, so I'll close this PR for now.

PR #1478, opened 2023-05-17
Suggest/Strings, 15.6.1 Encoding, inline CSVs don't work

mitsuoxv:
Inline CSVs including binary data are not recognized as CSVs in my environment (Linux).
As the following code chunks are eval = false, I would like to suggest to write x1 as a CSV file "x1.csv".

### DIFF
--- a/strings.qmd
+++ b/strings.qmd
@@ -545,7 +545,7 @@ readr uses UTF-8 everywhere.
 This is a good default but will fail for data produced by older systems that don't use UTF-8.
 If this happens, your strings will look weird when you print them.
 Sometimes just one or two characters might be messed up; other times, you'll get complete gibberish.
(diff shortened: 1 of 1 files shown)


mine-cetinkaya-rundel:
@hadley I'm actually running into new issues with reading inline CSVs. Would you try this one as well?

When I run

```r
x1 <- "text\nEl Ni\xf1o was particularly bad this year"
[read_csv](https://readr.tidyverse.org/reference/read_delim.html)(x1)$text
```

RStudio crashes.

jennybc:
Some context: https://github.com/tidyverse/readr/issues/1521

mine-cetinkaya-rundel:
Thanks @jennybc! I'll close this PR and implement the change suggested in the linked thread.

PR #1483, opened 2023-05-19
fix typo

tomjemmett:


mine-cetinkaya-rundel:
Thank you for the feedback! Data can be plural too, so I think it's ok to leave as is.

PR #1484, opened 2023-05-19
Update workflow-basics.qmd

dougnagie:
fixed sentence fragment at very end of chapter


mine-cetinkaya-rundel:
Thank you for the feedback! I think the sentence sounds ok as is so I'll leave it.

PR #1489, opened 2023-05-22
Suggest/functions, clarify the role of the last line in the body

mitsuoxv:
I would like to suggest to add a sentence to clarify the role of the last line in the body.


hadley:
I think the rule is a bit too complicated to summarise so succinctly — it's really the last expression, not the last line, and the return value can also come earlier with an explicit `return()`.

PR #1492, opened 2023-05-23
Suggest/iteration; 27.2.4 Column names; instead example

mitsuoxv:
It is natural to use coalesce, the same function as the previous code chunk.


mine-cetinkaya-rundel:
Thank you!

PR #1493, opened 2023-05-23
Suggest/iteration; 27.2.6 across() in functions; .groups = "drop"

mitsuoxv:
Better to follow the good practice in 26.3.3 Common use cases.


mine-cetinkaya-rundel:
Thank you!

PR #1494, opened 2023-05-23
Fix/base-R typos and Figure 28.1 pepper caption

mitsuoxv:
I correct Figure 28.1 caption, according to 1st edition, 20.5.3 Lists of condiments.


hadley:
Thanks; I'll going to update the figure to better align with the text momentarily.

PR #1502, opened 2023-05-25
Suggest/quarto; image sizing

mitsuoxv:
Image sizing part is harder to understand in 2e Quarto than in 1e Rmarkdown:

- How to set global chunk options varies by option in Quarto, while all can be set in knitr::opts_chunk$set() in Rmarkdown.
- layout-ncol hides out-width in 2e.

I would like to suggest to add YAML example, and explanation about the relation between layout-ncol and out-width.


quarto.qmd
```diff
@@ -441,17 +441,32 @@ The biggest challenge of graphics in Quarto is getting your figures the right si
...
 
--   Control the output size with `out-width` and set it to a percentage of the line width.
-    We suggest `out-width: "70%"` and `fig-align: center`.
+-   Control the output size with `out-width` and set it to a percentage of the body width of the output document.
+    We suggest to `out-width: "70%"` and `fig-align: center`.
+
     That gives plots room to breathe, without taking up too much space.
 
 -   To put multiple plots in a single row, set the `layout-ncol` to 2 for two plots, 3 for three plots, etc.
+    This effectively sets `out-width` to "50%" for two, "33%" for three, etc.
```
mine-cetinkaya-rundel, line 469:
```suggestion
    This effectively sets `out-width` to "50%" for each of your plots if `layout-ncol` is 2, "33%" if `layout-ncol` is 3, etc.
```

quarto.qmd
```diff
@@ -441,17 +441,32 @@ The biggest challenge of graphics in Quarto is getting your figures the right si
...
+title: "Example"
+format: 
+  html:
+    fig-width: 6
+    fig-align: center
+knitr:
+  opts_chunk:
+    fig-asp: 0.618
+    out.width: "70%" # out-width doesn't work for now
+```
```
mine-cetinkaya-rundel, line 457:
```suggestion
We recommend three of the five options:
```

mine-cetinkaya-rundel:
Thanks! I've removed the sample document YAML because I think the mismatched knitr options in the document YAML and chunk headers are confusing. This is currently the case for Quarto, but it might either change (that would be my hope, so I'll open an issue about it) or it will get better documented (this is already on the roadmap for Quarto docs). I'll merge in the rest of the changes in this PR.

PR #1505, opened 2023-05-25
Fix/quarto mostly typos

mitsuoxv:
- label_comma() is a scales function, and returns a function, not a value. comma() function is created in the previous chunk.
- Insert menu is in the visual editor. If you say Rstudio menu, I look for the first upper menu.
- better to stick to out-width: "70%" notation.
- "By default these dependencies ..." probably forgot to delete.
- !r causes an error. !expr works.

### DIFF
--- a/quarto.qmd
+++ b/quarto.qmd
@@ -429,7 +429,7 @@ comma(.12358124331)
 
 The figures in a Quarto document can be embedded (e.g., a PNG or JPEG file) or generated as a result of a code chunk.
 
-To embed an image from an external file, you can use the Insert menu in the visual editor in RStudio and select Figure / Image.
+To embed an image from an external file, you can use the Insert menu in the Visual Editor in RStudio and select Figure / Image.
 This will pop open a menu where you can browse to the image you want to insert as well as add alternative text or caption to it and adjust its size.
 In the visual editor you can also simply paste an image from your clipboard into your document and RStudio will place a copy of that image in your project folder.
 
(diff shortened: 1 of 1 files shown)


quarto.qmd
```diff
@@ -422,14 +422,14 @@ comma(.12358124331)
 2.  Download `diamond-sizes.qmd` from <https://github.com/hadley/r4ds/tree/main/quarto>.
     Add a section that describes the largest 20 diamonds, including a table that displays their most important attributes.
 
-3.  Modify `diamonds-sizes.qmd` to use `label_comma()` to produce nicely formatted output.
+3.  Modify `diamonds-sizes.qmd` to use `comma()` to produce nicely formatted output.
```
mine-cetinkaya-rundel, line 425:
```suggestion
3.  Modify `diamonds-sizes.qmd` to use `label_comma()` to produce nicely formatted output.
```

quarto.qmd
```diff
@@ -429,7 +429,7 @@ comma(.12358124331)
 
 The figures in a Quarto document can be embedded (e.g., a PNG or JPEG file) or generated as a result of a code chunk.
 
-To embed an image from an external file, you can use the Insert menu in RStudio and select Figure / Image.
+To embed an image from an external file, you can use the Insert menu in the visual editor in RStudio and select Figure / Image.
```
mine-cetinkaya-rundel, line 432:
```suggestion
To embed an image from an external file, you can use the Insert menu in the Visual Editor in RStudio and select Figure / Image.
```

mitsuoxv:
If you stick to label_comma(), it should be written as scales::label_comma()(), as label_comma() is a scales function, and returns a function, not a value.
If you think you have created label_comma() by hand in the previous chunk, the function name is not label_comma() but comma().

PR #1506, opened 2023-05-25
Update data-import.qmd

davidkane9:
I think you want "sprinkled," not "speckled." But maybe I am misreading . . .


mine-cetinkaya-rundel:
Thank you!

Issue #1374, opened 2023-03-16
Section 23.4.3 Code returns error


bfletcherjohnson:
Code in Section 23.4.3, 2nd code box returns error, nearly immediately.  I think the problem is that cells in the ISBN column are empty until row 83240, that cell contains the ISBN for the book in that row.  open_dataset() assigns ISBN as null based on the first few thousand rows; when subsequently write_dataset encounters an number in ISBN, it throws an error.

Code in 2nd code box

```r
seattle_csv |>
  group_by(CheckoutYear) |>
  write_dataset(path = pq_path, format = "parquet")
```

returns error as shown below:

```r
> seattle_csv |>
+   group_by(CheckoutYear) |>
+   write_dataset(path = pq_path, format = "parquet")
Error: Invalid: In CSV column #7: Row #83240: CSV conversion error to null: invalid value '9781504752848'
```

There seems to be an error in csv file, column #7, row #83240. I hesitate to open a 9GB csv file

I deleted the directory and file and downloaded the file a second time; I get the same error.
Code in the previous section 23.3 works as described, suggesting that seattle.csv is NOT corrupt.

When I open the csv file in excel (excel, of course, complains that it is too big), in row 83240, column H (which is column #8 not #7) there is the number 9781504752848.  Column H is "ISBN" and appears to be empty (NA) for nearly all rows.  However, I  see that row 150026, column H contains 9781844678433, row 297216 contains 9781459234451.  These numbers are the correct ISBN for the book in the row; it is just that for the vast majority of books an ISBN is not given.

You mention that open_dataset() scans a few thousand rows to figure out the structure of the dataset.  Unfortunately, for the first few thousand rows the ISBN column is indeed null.

I am running :
RStudio 2022.12.0+353 "Elsbeth Geranium" Release (7d165dcfc1b6d300eb247738db2c7076234f6ef0, 2022-12-03) for Windows
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) RStudio/2022.12.0+353 Chrome/102.0.5005.167 Electron/19.1.3 Safari/537.36

hadley:
You'll need the latest version of arrow.

bfletcherjohnson:
[...]

PR #1517, opened 2023-06-06
Update arrow chapter code to avoid errors

thisisnic:
As discussed in #1374

I did mention using `open_csv_dataset()` in the linked issue, but just using `open_dataset()` might be simpler in the short-term as it requires less explanation, given there's no `open_parquet_dataset()` function.


raffaem:
Can you squash the two commits into 1?

hadley:
Thanks!

PR #1522, opened 2023-06-16
Small format for column

Lextuga007:
species column name is missing back ticks in this reference


hadley:
Thank you!

Lextuga007:
Thank you for a great book!

PR #1533, opened 2023-07-16
Download Seattle library

raffaem:
Download Seattle library if it wasn't downloaded already


mine-cetinkaya-rundel:
@hadley Would you take a look at this one? I agree with the suggestion and it works, however I'm unable to re-render the chapter due to a different error:

```r
seattle_csv |>
  group_by(CheckoutYear) |>
  write_dataset(path = pq_path, format = "parquet")
```

gives 

```
Error:
! Invalid: In CSV column #7: Row #83240: CSV conversion error to null: invalid value '9781504752848'
```

We do need to re-render this chapter and update the freeze since the WIP status is still visible on the book website.

mine-cetinkaya-rundel:
Related to #1374

hadley:
Yeah, this seems fine to me.

mine-cetinkaya-rundel:
Thank you!

PR #1538, opened 2023-07-18
Update data-visualize.qmd

emhofer:
fixed small typo


mine-cetinkaya-rundel:
Thank you! Data can be plural so either works, therefore I'll leave this suggestion for now.

PR #1565, opened 2023-08-21
26.1 Functions: three -> four advantages

PashaMS:
Very small issue.
Text says three advantages then lists four.
I believe 1e listed three.


mine-cetinkaya-rundel:
Thank you!

PR #1571, opened 2023-08-27
Fixed typo 'come' -> 'comes'

dfenny:
A minor change to fix some bad grammar in chapter 2.


mine-cetinkaya-rundel:
Data can be plural or singular, so either would work here.

PR #1582, opened 2023-10-13
Update data-visualize.qmd

duncanjunegong:
no 'linewidth' parameter in geom_density() but 'size'


mine-cetinkaya-rundel:
Thanks for the feedback. Note that starting with ggplot2 3.4.0, using `size` aesthetic for lines was deprecated. So `linewidth` is the correct argument here.

PR #1598, opened 2023-11-13
Tidyselect operations `!` and `-` are subtly different

lionel-:
Closes https://github.com/tidyverse/dplyr/issues/6959

### DIFF
--- a/data-transform.qmd
+++ b/data-transform.qmd
@@ -347,7 +347,7 @@ In this situation, the first challenge is often just focusing on the variables y
       select(!year:day)
     ```
 
-    Historically this operation was done with `-` instead of `!`, so you're likely to see that in the wild. These two operators serve the same purpose but with subtle differences in behaviour. We recommend to consistently use `!` because it reads as "not" and combines well with `&` and `|`.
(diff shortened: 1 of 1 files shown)


data-transform.qmd
```diff
@@ -347,7 +347,7 @@ In this situation, the first challenge is often just focusing on the variables y
       select(!year:day)
     ```
 
-    You can also use `-` instead of `!` (and you're likely to see that in the wild); we recommend `!` because it reads as "not", and combines well with `&` and `|`.
+    Historically this operation was done with `-` instead of `!`, so you're likely to see that in the wild. These two operators serve the same purpose but with subtle differences in behaviour. We recommend to consistently use `!` because it reads as "not" and combines well with `&` and `|`.
```
mine-cetinkaya-rundel, line 350:
```suggestion
    Historically this operation was done with `-` instead of `!`, so you're likely to see that in the wild. These two operators serve the same purpose but with subtle differences in behavior. We recommend using `!` because it reads as "not" and combines well with `&` and `|`.
```

dholstius:
Here's an example of a not-so-subtle difference? The final line silently declines to return `foo` at all.

```r
select(mtcars, foo = cyl)
select(mtcars, foo = cyl, !cyl)
select(mtcars, foo = cyl, everything(), -cyl)
```

PR #1611, opened 2023-12-19
Update workflow-basics.qmd

piotr-kaczmarski:
Letter c in the sentence:

"You can **c**ombine multiple elements into a vector with c():"  has a markdown around which makes it bold in the final book output. 

I think this may be a typo. Thanks so much!


piotr-kaczmarski:
ok just got it after scrolling to seq that is also used as bold in sequences - all good :)

PR #1621, opened 2024-01-19
Update quarto.qmd

e-clin:
Replaced "a ridiculous degree of **accuracy**" with "a ridiculous degree of **precision**", as that is what the number of decimal places relates to.


mine-cetinkaya-rundel:
Thank you!

PR #1623, opened 2024-01-25
logicals.qmd: use better variable names in example

florisvdh:
The variable names of this example were copied from the previous example, where logical vectors were summarized with `any()` and `all()`. However in the present example the summary is with `sum()` and `mean()`, which was not reflected (updated) in the variable names.


mine-cetinkaya-rundel:
Thank you!

PR #1641, opened 2024-03-28
Update workflow-style.qmd, Fixing minor typo?

davidrsch:
Adding spaces between words, I don't if it was intended to be without spaces


mine-cetinkaya-rundel:
It was intended to be without spaces, to make a point about style 😄

PR #1645, opened 2024-04-10
joins.qmd: fix party-birthday matching condition in text (fixes #1610)

### LINKED ISSUE #1610: 19.5.3 Rolling joins - parties example
> "And for each employee we want to find the first party date that comes after (or on) their birthday. We can express that with a rolling join:"

```
employees |> 
  left_join(parties, join_by(closest(birthday >= party)))
```

The sentence should rather be:
"And for each employee we want to find the **last** party date that comes **before** (or on) their birthday. We can express that with a rolling join:"

Alternatively, the code could be updated as follows

```
employees |> 
  left_join(parties, join_by(closest(birthday <= party)))
```

if we want parties after birthdays as mentioned in the original sentence

florisvdh:
The updated text now reflects what has been intended, as confirmed from reading on. It's the implementation of the proposed solution in #1610.

Fixes #1610.

### DIFF
--- a/joins.qmd
+++ b/joins.qmd
@@ -819,7 +819,7 @@ employees <- tibble(
 employees
 ```
 
-And for each employee we want to find the first party date that comes after (or on) their birthday.
+And for each employee we want to find the last party date that comes before (or on) their birthday.
 We can express that with a rolling join:
 
 ```{r}


mine-cetinkaya-rundel:
Thank you!

PR #1647, opened 2024-04-17
Update iteration.qmd

davidkane9:
Not sure what "Vs" is supposed to mean. Perhaps "Versus?" In any case, "Compare with" seems better.


mine-cetinkaya-rundel:
Thank you!

PR #1650, opened 2024-04-19
Update databases.qmd

davidkane9:
Fixed title and removed italics, since "The Three-Valued Logic of SQL" is an article, not a book.


mine-cetinkaya-rundel:
Thank you!

PR #1653, opened 2024-05-09
Suggest/facets be renewed

mitsuoxv:
Use vars() instead of ~ to quote a variable name in facet_wrap() and facet_grid(), like in functions.qmd.


mine-cetinkaya-rundel:
Thank you for the suggestions, however we used the `~` notation since that's the recommended notation for faceting functions.

PR #1659, opened 2024-05-24
Suggest/data-import.qmd

mitsuoxv:
probably a typo, and no mention about fixing `student_id` column in the rest of the chapter.


mine-cetinkaya-rundel:
Thank you!

PR #1660, opened 2024-05-28
Suggest/layers.qmd shape descriptions in fig-alt, etc.

mitsuoxv:
How about changing fig-alt descriptions of 26 shapes to those matching character strings, which can be used to set "shape"?
Other suggestions are mostly typos and careless mistakes.


mine-cetinkaya-rundel:
Thank you!

PR #1661, opened 2024-05-30
Suggest/EDA.qmd merge two sections of unusual values

mitsuoxv:
How about merging two sections of "unusual values" into one?
Other suggestions are mainly typos and careless mistakes.


mine-cetinkaya-rundel:
Thank you for the suggestion! I'm going to leave this open for now and we'll revisit when we're working on more substantial updates to the book.

PR #1664, opened 2024-06-01
Fix/logicals.qmd and transform.qmd; correction of fig-alt, and typos

mitsuoxv:
correction of fig-alt, and probably typos


mine-cetinkaya-rundel:
Thank you!

PR #1666, opened 2024-06-03
Correct numeric precision comment

capnrefsmmat:
It's not correct to say R uses numbers with "a fixed number of decimal places"; that implies fixed-point arithmetic, but R uses floating-point arithmetic, where the number of decimal places depends on the scale of the variable. Numbers near 0 are represented with more decimal places than very large numbers.

It's fair to say the precision is finite, and that's the real problem, as the precision isn't enough to represent the numbers exactly (and can't be, for sqrt(2)).

### DIFF
--- a/logicals.qmd
+++ b/logicals.qmd
@@ -100,7 +100,7 @@ x == c(1, 2)
 ```
 
 What's going on?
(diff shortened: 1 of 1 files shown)


hadley:
I agree that this is not technically correct but it's close enough to get the point across to the average reader here.

PR #1675, opened 2024-07-09
Correct display of embracing operator

capnrefsmmat:
Quarto supports inline code using the backtick operator [1], using backticks and curly quotes to specify the engine. Because of this, `{{ }}` is apparently parsed as inline code and renders as `{ }`. An extra level of curly braces must be added to display the embracing operator in the text.

(I suppose this could also be considered a Quarto bug, since their parser is treating this as inline code when it clearly isn't. But this is an easy solution.)

[1]: https://quarto.org/docs/computations/inline-code.html


florisvdh:
Just noting that this would solve a rather critical bug in the functions chapter, where the embracing operator is currently explained as being `{}`. We also get a clash between single and double braces in the following paragraph:

afbeelding

mine-cetinkaya-rundel:
Thank you!

PR #1678, opened 2024-07-28
when you have groups of columns in across

anilyayavar:
when you have groups of columns that you want to compute with simultaneously, we may have an approach using glue specs and get()


mine-cetinkaya-rundel:
Thank you for the suggestion. I think the text reads well as is, so I'll close this PR.

Issue #1122, opened 2022-11-05
Ordered factors


elinw:
This line
https://github.com/hadley/r4ds/blob/main/factors.qmd#L416
says that ordered factors  " equal distance between levels"  but that is not the case, if they were equidistant that would be interval and could be represented by a number.  What makes them distinct is that while ordered there is no underlying common unit of measurement.  E.g. Strongly agree could be a lot farther from Agree than Agree is from neutral.

Also it has two periods at the end.

hadley:
Can you provide some evidence for that claim? Ordered factors use polygonal contrasts, which if I understand correctly, imply an equal spacing of the levels.

XiangyunHuang:
> Can you provide some evidence for that claim? 

Take [likert](https://github.com/jbryer/likert) and [Likert scale](https://en.wikipedia.org/wiki/Likert_scale) for an example.

hadley:
@XiangyunHuang how is that related to the ordered factors class in R?

hadley:
I'm happy to continue this discussion, but I'm going to close this issue for now since it's not clear to me that anything needs to change in R4DS.

XiangyunHuang:
@hadley  

1. `?factor` gives in **Details** section

    >  Ordered factors differ from factors only in their class, but methods and the model-fitting functions treat the two classes quite differently.
    
    This implies lots of different in statistical models, linear model is just a small example. 

1. R4DS gives in **16.6 Ordered factors** section

    > Ordered factors, created with [ordered()](https://rdrr.io/r/base/factor.html), imply a strict ordering and equal distance between levels

    In fact, **Ordered factors** doesn't imply equal distance between levels. Therefore, linear models use “polygonal contrasts” and other methods to handle **Ordered factors**.  In survey research, [Likert scale](https://en.wikipedia.org/wiki/Likert_scale) is widely used to scaling responses which collects ordered category data. More details about how to processing them, please see **Agresti, Alan. 2013. Categorical Data Analysis. 3rd . Hoboken, New Jersey: John Wiley & Sons, Inc.** (highly recommended)

leorjorge:
@hadley I Would like to chip in here as well, even if this is a closed issue. Ordered factors are most definitely not equal distance between levels, otherwise they'd behave exactly like a numeric linear variable. Their application, and all the methods developed to deal with ordered factors both as response and as predictor are based on the idea they have an order but the distances between levels can vary (potentially arbitrarily). 

[...]

PR #1686, opened 2024-09-24
correct ordered factor definition

leorjorge:
Following discussion in #1122 , clarifies that ordered factors do not behave like numbers.

### DIFF
--- a/factors.qmd
+++ b/factors.qmd
@@ -410,7 +410,7 @@ Read the documentation to learn about `fct_lump_min()` and `fct_lump_prop()` whi
 ## Ordered factors {#sec-ordered-factors}
 
 Before we go on, there's a special type of factor that needs to be mentioned briefly: ordered factors.
-Ordered factors, created with `ordered()`, imply a strict ordering and equal distance between levels: the first level is "less than" the second level by the same amount that the second level is "less than" the third level, and so on.
+Ordered factors, created with `ordered()`, imply a strict ordering between levels: the first level is "less than" the second level, the second level is "less than" the third level, and so on. This sounds like a numeric variable, but while the difference between 1 and 2 is the same as the difference between 2 and 3, for ordered factors the difference between levels is not defined, so it is employed when you know that there is a ranking between the levels but the exact differences between levels is unknown.
 You can recognize them when printing because they use `<` between the factor levels:
 
 ```{r}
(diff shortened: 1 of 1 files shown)


leorjorge:
To clarify, this PR attempts to correct the obvious errors: the definition and the typo. 

I kept the more opinionated part discouraging people from using ordered factors, even if I don't really agree with it and they are quite important at least in the social sciences (and ecology to a lesser extent). In my experience with ecological data, they're used less often than they should, with people just treating these as either numeric variables or regular factors to avoid the use of more complicated models.

factors.qmd
```diff
@@ -421,9 +421,9 @@ In practice, `ordered()` factors behave very similarly to regular factors.
...
--   If you use an ordered function in a linear model, it will use "polygonal contrasts". These are mildly useful, but you are unlikely to have heard of them unless you have a PhD in Statistics, and even then you probably don't routinely interpret them. If you want to learn more, we recommend `vignette("contrasts", package = "faux")` by Lisa DeBruine.
+-   If you use an ordered predictor in a linear model, it will use "polynomial contrasts". These are mildly useful, but you are unlikely to have heard of them unless you have a PhD in Statistics, and even then you probably don't routinely interpret them. If you want to learn more, we recommend `vignette("contrasts", package = "faux")` by Lisa DeBruine.
 
-Given the arguable utility of these differences, we don't generally recommend using ordered factors.
+Ordered categories are frequently used in the social sciences but not so common in other fields, so unless you are faced with a dataset that has clearly ranked categories we don't generally recommend using ordered factors.
```
hadley, line 426: What's the impact on analysis of using them? Is it that the social sciences are using packages that have more behaviour differences between factors and ordered factors?

leorjorge: Yeah, my understanding is that as social sciences have so many survey-based data in a ranked scale, they use more often ordinal regression and other similar methods to deal with ordered categories both as predictors and as responses. For response variables, there is quite a bit of literature on ordinal regressions. these usually employ cumulative logit distributions. Methods to deal with ordered factors as predictors in a better way than these polynomial contrasts do seem to be much more recent, but are getting increasing use, at least in Bayesian analyses (se https://bpspsychub.onlinelibrary.wiley.com/doi/10.1111/bmsp.12195 for example).

hadley:
Thanks!

PR #1690, opened 2024-11-14
pull request

emunkhtsetseg:
I would like to translate the book.


mine-cetinkaya-rundel:
If you would like to translate the book, please send an email to hadley@posit.co with the language you're proposing to translate to. Thank you!

PR #1692, opened 2024-11-18
Adding Light and Dark mode

joshuamarie:
It would be great if the R4DS web book has also dark mode in it.


andrewpking:
Bump for accessibility!

PR #1701, opened 2025-01-11
Fix some typos

kleintom:
Thanks for making this available! Feel free to ask for any followup adjustments.

### DIFF
--- a/spreadsheets.qmd
+++ b/spreadsheets.qmd
@@ -347,7 +347,7 @@ You can write data back to disk as an Excel file using the `write_xlsx()` functi
 ```{r}
 #| eval: false
 
-write_xlsx(bake_sale, path = "data/bake_sale.xlsx")
+write_xlsx(bake_sale, path = "data/bake-sale.xlsx")
 ```
 
 @fig-bake-sale-excel shows what the data looks like in Excel.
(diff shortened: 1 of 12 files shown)


spreadsheets.qmd
```diff
@@ -342,12 +342,12 @@ bake_sale <- tibble(
...
 ```
 
-You can write data back to disk as an Excel file using the `write_xlsx()` from the [writexl package](https://docs.ropensci.org/writexl/):
+You can write data back to disk as an Excel file using the `write_xlsx()` function from the [writexl package](https://docs.ropensci.org/writexl/):
 
 ```{r}
 #| eval: false
 
-write_xlsx(bake_sale, path = "data/bake-sale.xlsx")
+write_xlsx(bake_sale, path = "data/bake_sale.xlsx")
```
kleintom, line 350: This matches the name in the caption text for the associated `Figure 20.4: Spreadsheet called bake_sale.xlsx in Excel.`: `bake_sale` with an underscore.

mine-cetinkaya-rundel (changes requested):
Thank you for these edits! I left one comment but I'm happy to make that change too if you prefer.

mine-cetinkaya-rundel (approved):

mine-cetinkaya-rundel:
Thank you!

PR #1740, opened 2025-10-01
Update workflow-basics.qmd

betahat:
Just fixes a typo


mine-cetinkaya-rundel:
Thank you! However this is an intentional mistake, since the exercise asks to fix errors, so I'll close this PR.

PR #1743, opened 2025-10-14
Clarify how to override the default type-guessing heuristic in data-i…

gabrivera:
…mport.qmd

The phrase took time to understand, and —unless I missed something about the "default column"—, I believe the idea here is that the _default algorithm_ for type-guessing is the one that's overridden by the "cols(.default" function and parameter.


mine-cetinkaya-rundel:
Thank you!

PR #1783, opened 2026-04-08
Add Vietnamese translation link

quan-nk:
A Vietnamese translation is now available at https://vn.r4ds.hadley.nz/


hadley:
Thank you!

